PR #6, opened 2019-08-12
CRD Strategy: implement concurrencyPolicy

### LINKED ISSUE #1: CRD Strategy: may fail when multiple spec updates happen during upgrade
Scenario:
- Create an InstanceGroup with CRD strategy submitting a workflow for upgrade.
- Update the InstanceGroup AMI, let node rotation start
- Mid-node rotation, change AMI again

This seems to break things - we should have some sort of locking mechanism or come up with some approach that accommodates the continuously converging nature of the controller vs. submitting workflows which are 'oneshot' or job style execution.

eytan-avisror:
Fixes #1 

- This implements concurrencyPolicy to lock currently running workflows/CR.
- Added some doc/example fixes

## Testing done
- Unit tests passing
- Full bdd test testing
- Manually ran rollingupgrades with InstanceGroup AMI changes and only a single CR was processed at a time.

### DIFF
--- a/controllers/provisioners/ekscloudformation/helpers.go
+++ b/controllers/provisioners/ekscloudformation/helpers.go
@@ -75,12 +75,9 @@ func getUnstructuredPath(u *unstructured.Unstructured, jsonPath string) (string,
 	}
 	statusPath := strings.FieldsFunc(jsonPath, splitFunction)
 
-	value, ok, err := unstructured.NestedString(u.UnstructuredContent(), statusPath...)
+	value, _, err := unstructured.NestedString(u.UnstructuredContent(), statusPath...)
 	if err != nil {
 		return "", err
 	}
-	if ok {
-		return value, nil
-	}
-	return "", nil
+	return value, nil
 }
--- a/controllers/provisioners/ekscloudformation/helpers.go
+++ b/controllers/provisioners/ekscloudformation/helpers.go
@@ -2,10 +2,12 @@ package ekscloudformation
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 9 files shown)


controllers/provisioners/ekscloudformation/helpers.go
```diff
@@ -47,3 +49,38 @@ func getConfigMap(k kubernetes.Interface, namespace string, name string, options
...
+		return c == '.'
+	}
+	statusPath := strings.FieldsFunc(jsonPath, splitFunction)
+
+	value, ok, err := unstructured.NestedString(u.UnstructuredContent(), statusPath...)
+	if err != nil {
+		return "", err
+	}
+	if ok {
+		return value, nil
```
kevdowney: What's in `value` if `ok` is false? You could probably just return value here if it's already empty.
Another improvement for later is to possibly return here `interface{}` type and reflect.Type the call to `Nested{Type}` function.

controllers/provisioners/ekscloudformation/strategy.go
```diff
@@ -68,8 +126,14 @@ func (ctx *EksCfInstanceGroupContext) processCRDStrategy() error {
 		return err
 	}
 
+	addAnnotation(customResource, OwnershipAnnotationKey, "true")
```
kevdowney: If this annotation is meant to convey "ownership" by another resource or your controller might be better to use a Kubernetes ownerreference instead. https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

If this is meant to convey some management association you could use a best-practice label here e.g., https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
`app.kubernetes.io/managed-by: instancemgr.orkaproj.io`

kevdowney: You might want both annotation and label here depending on your use-case.

eytan-avisror: since CRD strategy is generic, I don't know what CR the user will submit as part of upgrade, hence I cannot foresee if `ownerReference` will be supported by the controller.
I will look into changing the annotation key/value to best practice

PR #28, opened 2019-08-20
Retain existing aws-auth mapRoles

### LINKED ISSUE #26: instance-manager strips out user permissions from aws-auth configmap
Looks like instance-manager is stripping out user permissions from aws-auth configmap.
For example, given a configmap:
```
apiVersion: v1
data:
  mapRoles: |
    - rolearn: arn:aws:iam::1234567890:role/a-valid-node-role
      username: system:node:{{EC2PrivateDNSName}}
      groups:
        - system:bootstrappers
        - system:nodes
    - rolearn: arn:aws:iam::1234567890:role/an-invalid-node-role
      username: system:node:{{EC2PrivateDNSName}}
      groups:
        - system:bootstrappers
        - system:nodes
    - rolearn: arn:aws:iam::1234567890:role/a-valid-user
      username:  user:some-user
      groups:
        - system:masters
kind: ConfigMap
metadata:
  annotations:
  name: aws-auth
  namespace: kube-system
```

After reconciling, instance-manager will strip out last two mapRoles.
We should have some mechanism of **NOT** stripping out user access, possibly we can look at `.username` if it starts with `user:` and as a result not touch it.

eytan-avisror:
Fixes #26 

- Parsing of the existing `aws-auth` configmap
- Retain mapRole if it does not conflict with a node role

## Testing
-- Added some unit tests

### DIFF
--- a/controllers/provisioners/ekscloudformation/bootstrap.go
+++ b/controllers/provisioners/ekscloudformation/bootstrap.go
@@ -18,9 +18,12 @@ package ekscloudformation
 import (
 	"sort"
 
+	"github.com/orkaproj/instance-manager/api/v1alpha1"
 	"github.com/orkaproj/instance-manager/controllers/common"
+
 	yaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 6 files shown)


controllers/provisioners/ekscloudformation/bootstrap.go
```diff
@@ -44,9 +44,26 @@ func (ctx *EksCfInstanceGroupContext) getActiveNodeArns() []string {
...
+	existingConfigurations := []AwsAuthConfig{}
+	newConfigurations := []AwsAuthConfig{}
+
+	err := yaml.Unmarshal([]byte(existingAuthMap.Data["mapRoles"]), &existingConfigurations)
+	if err != nil {
+		return err
+	}
+
+	for _, configuration := range existingConfigurations {
+		// if existing config is not part of discovered node ARNs, add it as is
```
tekenstam: Any reason not to do an upsert of the existingConfigurations? When an IG is deleted, is it removed from the config map?

eytan-avisror: @tekenstam 
Yes, I think I need to clarify the code a little bit + handle deletions now that we are no longer constructing the configmap from scratch.

Will re-iterate and add a commit soon

controllers/provisioners/ekscloudformation/types.go
```diff
@@ -34,6 +36,29 @@ type AwsAuthConfigMapRolesData struct {
...
+	}
+	m.MapRoles = append(m.MapRoles, config)
+}
+
+func (m *AwsAuthConfigMapRolesData) Remove(config AwsAuthConfig) {
+	for i := len(m.MapRoles) - 1; i >= 0; i-- {
+		role := m.MapRoles[i]
+
+		if reflect.DeepEqual(role, config) {
+			m.MapRoles = append(m.MapRoles[:i],
```
tekenstam: Is this function used? I'm not following how the config is removed by `append`.

eytan-avisror: It removes from the list by index (using append) - see `re-slicing`
But good catch - looks like it's not needed and is a remnant if previous commit, removal is now handled by not discovering deleting node groups.

I'll remove this method

PR #46, opened 2019-09-13
Cloudformation Stack Prefixes

### LINKED ISSUE #45: Add configurable prefix to stack name
The current stack name is constructed as:
CLUSTERNAME-NAMESPACE-GROUPNAME

We should change it to:

<CONFIGURABLE_PREFIX>-CLUSTERNAME-GROUPNAME

- Configurable prefix can be added to the controller configmap as an option `stackNamePrefix`.
- Constructed name must be < 63 characters or operation should fail.

eytan-avisror:
Fixes #45 

This PR adds support for `stackNamePrefix` in the controller global config.
Created stack name is persisted in `.Status.StackName` and then only read from there.
Move `LoadControllerConfig` inside the provisioner as it's quite specific for the ekscf provisioner and that seem to make more sense inside rather then in the controller.

Future PRs will follow for documentation improvements and improvements to current unit tests which are becoming a bit brittle.

Tested this feature out manually with various scenarios, including changing the prefix and restarting the controller, everything works as expected since the stack name is now persisted in the status and is the source of truth for stack name

### DIFF
--- a/api/v1alpha1/instancegroup_types.go
+++ b/api/v1alpha1/instancegroup_types.go
@@ -177,6 +177,7 @@ type EKSCFConfiguration struct {
 
 // InstanceGroupStatus defines the schema of resource Status
 type InstanceGroupStatus struct {
+	StackName                     string `json:"stackName,omitempty"`
 	CurrentState                  string `json:"currentState,omitempty"`
 	CurrentMin                    int    `json:"currentMin,omitempty"`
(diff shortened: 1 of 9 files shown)


eytan-avisror:
> LGTM. If you're changing the name of the status field it is a breaking change so you might want to consider documenting that.

This doesnt change existing status fields, simply introduce a new one `StackName`

Issue #53, opened 2019-09-24
Support for MetricsCollection


eytan-avisror:
Currently spec does not allow enabling MetricsCollection.
We should probably add support for it.

More info in the API docs:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-metricscollection.html

RiyaJohn:
Can I try this issue?

eytan-avisror:
Hi @RiyaJohn 
Sure! happy to take contributions.

I think #54 is very similar but is more urgently needed.

In any case feel free to work on either, I will add some more details to both issues

eytan-avisror:
Implementation notes:

- We need to extend the InstanceGroup API to take `.spec.eks-cf.configuration.metrics` which is a list of metrics which should be added to the cloudformation template under a new object.

- In the current instance-manager config map that is in `config/crd/bases/instance-manager-configmap.yaml` there is a config map that holds the cloudformation template being used. we should modify the section that creates the autoscaling group and add templating for the MetricsCollection.

```
NodeGroup:
        Type: AWS::AutoScaling::AutoScalingGroup
        Properties:
          ...
          ...
          <we should add the MetricsCollection object here via go templating, for example>
          MetricsCollection:
            Metrics:
              - <user-requested-metric-1>
              - <user-requested-metric-1>
            Granularity: "1Minute" >> this is the only allowed value - we can hard code it
```

- Make sure to also take your changes to `docs/04_instance-manager.yaml` which has a copy of the configmap

- It will be nice to check whether provided metrics are one of allowed values and fail if it's not, however not sure current structure has a good place to add this in, we can do this with this PR or a future PR.

- Another nice-to-have is support for metric called `all` which would simply add all of the supported metrics

Essentially a user will provide input via the CR like so:
```
apiVersion: instancemgr.keikoproj.io/v1alpha1
kind: InstanceGroup
metadata:
  name: hello-world
spec:
  provisioner: eks-cf
  strategy:
    ...
  eks-cf:
    ...
    configuration:
      ...
      metrics:
        - groupMinSize
        - groupMaxSize
        - groupDesiredCapacity
        - groupInServiceInstances
        - groupPendingInstances
        - groupStandbyInstances
        - groupTerminatingInstances
        - groupTotalInstances
```

Which would change the cloudformation template to:

```
...
NodeGroup:
        Type: AWS::AutoScaling::AutoScalingGroup
        Properties:
          ...
          ...
[...]

PR #55, opened 2019-10-03
updating with managed policies logic

mnkg561:
This addresses https://github.com/keikoproj/instance-manager/issues/54 

There is a slight change in implementation as mentioned in the issue, This proposed solution expects user to provide ManagedPolicyName instead of Managed Policy ARN to avoid any syntax issues in ARN. This solution takes the names and construct the ARNs.

PR #57, opened 2019-10-15
Support for MetricsCollection

RiyaJohn:
Fixing  #53 Developed and tested support for metrics collection with metric option -all which would simply add all of the supported metrics.
Testing done:
Added TestGetNodeAutoScalingGroupMetrics in ekscloudformation_test.go that covers
case 1 when metricsCollection section is not specified at all
case 2 for metricsCollection:
                    - groupMinSize
                    - groupMaxSize
case 3 for metricsCollection:
                    - all
I had used sigs.k8s.io/controller-tools **v0.2.1** as this sigs.k8s.io/controller-tools/cmd/controller-gen **@v0.2.0-beta.2** was throwing error while I tried setting up, so there are many changes in go.sum accordingly, should I commit them too?


RiyaJohn:
By mistake, my initial commit got a wrong author name rjohn3. Can the cla block be skipped for rjohn3? Else I will have to clone repo afresh again and commit changes?

docs/04_instance-manager.yaml
```diff
@@ -843,6 +851,17 @@ data:
           - Key: KubernetesCluster
             Value: !Ref ClusterName
             PropagateAtLaunch: 'true'
+          {{if .Spec.EKSCFSpec.EKSCFConfiguration.MetricsCollection}}
```
eytan-avisror: It looks like we can avoid templating entirely if we resolve 'all' from inside `getNodeAutoScalingGroupMetrics()` function.
I think if we can avoid templating it would be a cleaner solution.

So, if you can resolve `all` from inside `getNodeAutoScalingGroupMetrics()`, it means you can just pass a parameter to the cloudformation template and then this block becomes:
```
MetricsCollection:
          - Metrics:
            !Ref NodeAutoScalingGroupMetrics
            Granularity: "1Minute"
```

eytan-avisror:
Thanks @RiyaJohn it looks great.
I just had one comment above that I think we can improve the resolving of `all` and avoid go-templating if possible.

Can you also add a `testing done` section and elaborate a bit on the scenarios you tested?

Regarding the CLA - afraid it is not possible the skip this check, can you sign in with the other user and try to sign it? If you look at the last comment of #55 I think he had the same issue.

Regarding the change for `sigs.k8s.io/controller-tools` - we can possibly do this in a separate PR - we should probably run a full e2e test before merging it too.

You can run a full BDD test on your changes if you'd like (see https://github.com/keikoproj/instance-manager/blob/master/.github/DEVELOPER.md#running-bdd-tests)

Thanks again for the PR

docs/04_instance-manager.yaml
RiyaJohn: We can do that, but that will become hard-coding the list of these Allowed Values:
GroupMinSize
GroupMaxSize
GroupDesiredCapacity
GroupInServiceInstances
GroupPendingInstances
GroupStandbyInstances
GroupTerminatingInstances
GroupTotalInstances
Then later if AWS adds anything new then, we will have to add to code right.

eytan-avisror: I see, so if we do not specify `Metrics:` cloudformation defaults to all metrics.
This makes sense in that case.

RiyaJohn:
> Thanks @RiyaJohn it looks great.
> I just had one comment above that I think we can improve the resolving of `all` and avoid go-templating if possible.
> 
> Can you also add a `testing done` section to the PR description and elaborate a bit on the scenarios you tested?
> 
> Regarding the CLA - afraid it is not possible the skip this check, can you sign in with the other user and try to sign it? If you look at the last comment of #55 I think he had the same issue.
> 
> Regarding the change for `sigs.k8s.io/controller-tools` - we can possibly do this in a separate PR - we should probably run a full e2e test before merging it too.
> 
> You can run a full BDD test on your changes if you'd like (see https://github.com/keikoproj/instance-manager/blob/master/.github/DEVELOPER.md#running-bdd-tests)
> 
> Thanks again for the PR

Added testing done section to PR. Fixed the cla issue, Thanks to the comment in #55, it helped.

docs/04_instance-manager.yaml
RiyaJohn: Yes

Issue #47, opened 2019-09-13
Improve/Re-Write unit-tests


eytan-avisror:
Current unit tests is becoming brittle.. We should have a better design for test unit tests (internal vs. external) and have better structure for unit tests so that they don't break easily and it's easy to extend them.
This should also include some refactoring to make code more testable
We should also consider other packages like the controller itself, etc. which are currently not covered.

eytan-avisror:
closing this since we will be deprecating eks-cf provisioner in favor of new one.

PR #58, opened 2019-10-18
added unit test

forThePurposeOfCloning:
Contributes to #47

### DIFF
--- a/controllers/provisioners/ekscloudformation/ekscloudformation_test.go
+++ b/controllers/provisioners/ekscloudformation/ekscloudformation_test.go
@@ -1272,12 +1272,11 @@ func TestGetNodeAutoScalingGroupMetrics(t *testing.T) {
 	}
 }
 
-func TestParseCustomResourceYaml(t *testing.T) {
+func TestParseCustomResourceYamlEmptyString(t *testing.T) {
 	empty, err := common.ParseCustomResourceYaml("")
 	if (len(empty.Object) > 0)  {
-		t.Fatal("Object should be empty")
-		t.Fail();
-
+		t.Fatalf("Expected ParseCustomResourceYaml to return Unstructured whose Unstructured.object is of length: 0 but" +
+			" instead object is of length %d", len(empty.Object));
 	}
 	if err != nil {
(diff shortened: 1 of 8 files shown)


forThePurposeOfCloning:
@eytan-avisror please review

controllers/provisioners/ekscloudformation/ekscloudformation_test.go
```diff
@@ -1272,3 +1272,14 @@ func TestGetNodeAutoScalingGroupMetrics(t *testing.T) {
 	}
 }
 
+func TestParseCustomResourceYaml(t *testing.T) {
+	empty, err := common.ParseCustomResourceYaml("")
+	if (len(empty.Object) > 0)  {
+		t.Fatal("Object should be empty")
+		t.Fail();
```
eytan-avisror: I think `t.Fail()` is not needed here if you already do `t.Fatal()`
Also can you add a more descriptive message (for both `t.Fatal()` calls) that includes `expected: <>, got: <>`

controllers/provisioners/ekscloudformation/ekscloudformation_test.go
```diff
@@ -1272,3 +1272,14 @@ func TestGetNodeAutoScalingGroupMetrics(t *testing.T) {
 	}
 }
 
+func TestParseCustomResourceYaml(t *testing.T) {
```
eytan-avisror: Is the goal of this test to verify `ParseCustomResourceYaml` can handle an empty string?
Maybe we should rename the test to `TestParseCustomResourceYamlEmptyString`

test-bdd/main_test.go
```diff
@@ -76,6 +76,8 @@ func TestE2e(t *testing.T) {
 	RunSpecsWithDefaultAndCustomReporters(t, "InstanceGroup Type Suite", []Reporter{junitReporter})
 }
 
+
+
```
eytan-avisror: seems like we can undo these extra blank lines

forThePurposeOfCloning: Quick question. What files besides the .idea files would you want removed?

Issue #70, opened 2019-12-09
EPIC: Fargate Support 


thomaswhitcomb:
**Is this a BUG REPORT or FEATURE REQUEST?**:

Feature Request

**What happened**:

Amazon announced Fargate support for EKS.  Now pods can be scheduled via Fargate.

**What you expected to happen**:

Pods within a namespace (and optional tags) can be automatically scheduled to run via Fargate.  Users should be able to create, modify and delete Fargate support.

**Anything else we need to know?**:

https://docs.aws.amazon.com/eks/latest/userguide/fargate.html
https://docs.aws.amazon.com/eks/latest/userguide/fargate-profile.html

edlee2121:
We should integrate this with instance manager on IKS2. That is, a node group could be implemented using Fargate. Please keep this issue updated as we learn now to best integrate Fargate with IKS2.

thomaswhitcomb:
Yeah, I created an issue in https://github.com/keikoproj/instance-manager
and was thinking it will be implemented as another provisioner
(eks-fargate).  I am working on the spec now and will create an early PR
for the readme before i start writing code.

eytan-avisror:
Ref #71 

Once this initial implementation is in we should start opening issues and link them here.

eytan-avisror:
Closing this for now as we have basic fargate support, we can create more stories in the future should we want to improve the existing implementation

PR #71, opened 2019-12-19
Eks fargate

thomaswhitcomb:
Initial implementation of #70 

@eytan-avisror Love to have your thoughts and comments.


README.md
```diff
@@ -78,6 +80,51 @@ spec:
...
+metadata:
+  name: hello-world
+spec:
+  # provision for EKS using Fargate
+  provisioner: eks-fargate
+  # provisioner configuration
+  eks-fargate:
+    fargateProfileName: "my-unique-profile-name"
+    clusterName: "the-cluster-for-my-pods"
+    podExecutionRoleArn: "arn:aws:iam::123456789012:role/MyPodRole"
```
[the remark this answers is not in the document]
thomaswhitcomb: (1) Yup, can definitely create the execution role for the caller.  Most likely they will would need to modify it for their pod's specific resource needs.  Would you imagine surfacing the role in iksm?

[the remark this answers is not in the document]
thomaswhitcomb: (2) I was going with the longer name since it mapped directly to the AWS cloudformation definition.  Figured it would eliminate some indirection while debugging.  But, I can do either way.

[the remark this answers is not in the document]
eytan-avisror: (1) I think if an existing role is not provided we should try to create a very basic one that only grants the bare minimum required to schedule pods or use the basic functionality of fargate. Otherwise if people want a custom role or don't want to have their controller creating roles, they can provide an existing role which they pre-create. IKSM is not relevant here because this is open source :)

(2) Leaving this choice up to you 👍 fine with both

README.md
```diff
@@ -78,6 +80,51 @@ spec:
...
+    subnets:
+    - subnet-1a2b3c4d
+    - subnet-4d3c2b1a
+    - subnet-0w9x8y7z
+    # Unique, case-sensitive identifier that you provide to ensure the idempotency of the request
+    clientRequestToken: ""
```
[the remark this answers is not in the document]
thomaswhitcomb: It is an optional field, that permits the caller to issue the same request and get the same response.  For example, a caller could make multiple requests to create the same fargate profile and be guaranteed to get the same result.   HTTP DELETE and PUT are other example of idempotent requests.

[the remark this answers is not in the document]
eytan-avisror: I see.. in that case I don't think we need to expose this in the API, do we?
Do you expect anyone to be submitting a CR with this field? what would that mean?
This makes more sense maybe for internal logic when you want to make sure you are not creating something twice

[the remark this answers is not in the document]
thomaswhitcomb: @eytan-avisror I am going to write up a separate document on the use, design and approach.  I'll link to it from the README.md.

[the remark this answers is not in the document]
thomaswhitcomb: Changed my mind on the bit about a separate doc.  I wrote it up and there just wasn't much to it.  I'll add the pertinent bits to the README.md and we can decide later if a separate doc is worth it.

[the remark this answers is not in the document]
thomaswhitcomb: I agree with the suggestion on the `clientRequestToken`.  I removed it from the spec.

api/v1alpha1/instancegroup_types.go
```diff
@@ -42,6 +42,14 @@ const (
...
+type InstanceGroupSpec struct {
+	Provisioner        string             `json:"provisioner"`
+	EKSCFSpec          *EKSCFSpec         `json:"eks-cf,omitempty"`
+	EKSFargateSpec     *EKSFargateSpec    `json:"eks-fargate,omitempty"`
+	AwsUpgradeStrategy AwsUpgradeStrategy `json:"strategy"`
+}
+
 // InstanceGroup is the Schema for the instancegroups API
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:path=instancegroups,scope=Namespaced,shortName=ig
```
eytan-avisror: These annotations need to remain above `InstanceGroupSpec`

thomaswhitcomb: ok

edlee2121:
@mukulikak

eytan-avisror:
Some comments:
- You are not using the shared interface `HandleReconcileRequest` and instead implemented your own `HandleRequest` - you should use the common provisioner interface `CloudDeployer`

- AWS clients are created in the controller and not in the method call, could you refactor to use same pattern? example:
https://github.com/keikoproj/instance-manager/blob/master/controllers/instancegroup_controller.go#L303-L309

thomaswhitcomb:
@eytan-avisror Thanks for those comments.  I'll first resolve the conflicts then I'll fix up the first two points.

thomaswhitcomb:
@eytan-avisror Working on the first 2 comments above... Are we striving to encapsulate all AWS references to the controllers/providers/aws/aws.go?

eytan-avisror:
> @eytan-avisror Working on the first 2 comments above... Are we striving to encapsulate all AWS references to the controllers/providers/aws/aws.go?

All the AWS specific calls should be methods under object `AwsWorker`, which contains all required clients - this makes unit testing easier because we can pass in fake clients when we instantiate AwsWorker.

As far as order/organization, perhaps down the road (not in this PR), we can refactor to have files according to client (e.g. eks.go, cloudformation.go, etc).

config/crd/bases/instancemgr.keikoproj.io_instancegroups.yaml
```diff
@@ -3,8 +3,6 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  annotations:
-    controller-gen.kubebuilder.io/version: v0.2.4
```
eytan-avisror: Can you update to controller-gen 0.2.4 (go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4) and then run `make manifests`, you are using an older controller-gen which tries to add back all this stuff :)

thomaswhitcomb: done

eytan-avisror: after running `make manifests` it should modify these files back (CRD + role.yaml) to the minimal version, which you can check in to revert these changes

thomaswhitcomb:
Using ```HandleReconcileRequest``` and have refactored to the pattern shown here: https://github.com/keikoproj/instance-manager/blob/master/controllers/instancegroup_controller.go#L303-L309

docs/04_instance-manager.yaml
```diff
@@ -665,7 +703,8 @@ spec:
         app: instance-manager
     spec:
       containers:
-      - image: keikoproj/instance-manager:master
+      # image: keikoproj/instance-manager:master
+      - image: givt002/instance-manager:latest
```
eytan-avisror, line 707: probably should undo this one

README.md
```diff
@@ -80,6 +82,92 @@ spec:
         - key: mykey
           value: myvalue
 ```
+#### *eks-tf* provisioner
```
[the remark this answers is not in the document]
thomaswhitcomb: done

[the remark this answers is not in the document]
thomaswhitcomb: Also, did you want to remove line 38,29 in the table?

README.md
```diff
@@ -80,6 +82,92 @@ spec:
...
+#   - PolicyName: root
+#     PolicyDocument:
+#       Version: 2012-10-17
+#       Statement:
+#         - Effect: Allow
+#           Action: '*'
+#           Resource: '*'
+
+```
+Most likely an execution role with access to addtional AWS resources will be required.  In this case, the above IAM role can be used as the basis to create a new, custom role. Uncomment the remaining lines in the above role to add custom policies, create the new role (via the AWS CLI or console) and then rerun the **eks-fargate** provisoner.  Specify the custom role ARN as the `podExecutionRoleArn` parameter value.
```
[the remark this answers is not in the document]
thomaswhitcomb: Could have been written better but it meant that once you create the new role/policies, you need to specify the new role's ARN in the fargate spec and rerun the create.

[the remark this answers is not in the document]
thomaswhitcomb: i will reword it.

README.md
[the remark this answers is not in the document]
eytan-avisror: Sure, you can remove those as well for now.

README.md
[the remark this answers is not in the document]
eytan-avisror: If you modify an existing policy (the one you already provided the CR with) there should be no action required (it's already in effect), if you create a new role you need to delete and re-create the CR.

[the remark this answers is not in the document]
thomaswhitcomb: correct

README.md
```diff
@@ -80,6 +82,92 @@ spec:
...
+#           Action: '*'
+#           Resource: '*'
+
+```
+Most likely an execution role with access to addtional AWS resources will be required.  In this case, the above IAM role can be used as the basis to create a new, custom role. Uncomment the remaining lines in the above role to add custom policies, create the new role (via the AWS CLI or console) and then rerun the **eks-fargate** provisoner.  Specify the custom role ARN as the `podExecutionRoleArn` parameter value. 
+
+The **eks-fargate** provisioner does not have any unique state.  Using the AWS API, the provisoner will access the requested/targeted EKS cluster and Fargate Profile.  In addition to any AWS errors that occur, error status codes will be logged and returned on a request to:
+
+- Create a profile under an existing name.
+- Delete a non-existing profile.
```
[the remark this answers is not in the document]
thomaswhitcomb: yup

controllers/instancegroup_controller.go
```diff
@@ -377,6 +416,7 @@ func (r *InstanceGroupReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 	log.Infof("resource provisioner: %v", strings.ToLower(ig.Spec.Provisioner))
 
 	// Add Finalizer if not present, and set the initial state
+	// Provisioner EKS-Cloudformation Reconciler
```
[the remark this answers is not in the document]
thomaswhitcomb: done

controllers/providers/aws/aws.go
```diff
@@ -507,3 +533,278 @@ func IsStackInConditionState(key string, condition string) bool {
...
+	RoleExists bool
+	Profile    *eks.FargateProfile
+}
+
+func (state *ResourceState) GetRoleExists() bool {
+	return state.RoleExists
+
+}
+func (state *ResourceState) GetProfileState() *string {
+	none := "NONE"
```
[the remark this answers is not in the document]
thomaswhitcomb: will do.

controllers/providers/aws/aws.go
```diff
@@ -507,3 +533,278 @@ func IsStackInConditionState(key string, condition string) bool {
...
+	if state.Profile != nil {
+		return state.Profile.Status
+
+	} else {
+		return &none
+	}
+}
+func (state *ResourceState) IsProvisioned() bool {
+	return *state.GetProfileState() != "NONE"
```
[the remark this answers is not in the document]
thomaswhitcomb: done

controllers/providers/aws/aws.go
```diff
@@ -507,3 +533,278 @@ func IsStackInConditionState(key string, condition string) bool {
...
+		ClusterName:         w.ClusterName,
+		FargateProfileName:  w.ProfileName,
+		PodExecutionRoleArn: w.ExecutionArn,
+		Selectors:           w.Selectors,
+	}
+	_, err := w.EksClient.CreateFargateProfile(input)
+	return err
+}
+
+var defaultPolicyArn = "arn:aws:iam::aws:policy/AmazonEKSFargatePodExecutionRolePolicy"
```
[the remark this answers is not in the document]
thomaswhitcomb: yup

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,378 @@
...
+}
+func (ctx *InstanceGroupContext) CloudDiscovery() error {
+	return nil
+}
+func (ctx *InstanceGroupContext) Delete() error {
+	var err error
+	worker := ctx.AwsFargateWorker
+	instanceGroup := ctx.GetInstanceGroup()
+	if instanceGroup.Status.GetFargateRoleName() != "" {
+		err = worker.DeleteDefaultRolePolicy()
```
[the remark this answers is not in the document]
thomaswhitcomb: I think it is ok.  Status.SetFargateRoleName() is only called if we create the role.

controllers/providers/aws/aws.go
```diff
@@ -507,3 +533,278 @@ func IsStackInConditionState(key string, condition string) bool {
...
+		//profile exists
+		return nil
+	}
+
+	for i := 0; i < w.RetryLimit && err != nil; i++ {
+		log.Infof("CreateProfile - %d try - creating profile for cluster: %s, name: %s, arn: %s", i, *w.ClusterName, *w.ProfileName, *arn)
+		time.Sleep(time.Duration(i*500) * time.Millisecond)
+		_, err = w.EksClient.CreateFargateProfile(input)
+	}
+	return err
```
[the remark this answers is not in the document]
thomaswhitcomb: I put the retry in because I almost always experienced race conditions between creating the role and then using the role in the fargate profile creation.

controllers/providers/aws/aws.go
```diff
@@ -507,3 +533,278 @@ func IsStackInConditionState(key string, condition string) bool {
...
+		ClusterName:        w.ClusterName,
+		FargateProfileName: w.ProfileName,
+	}
+	output, err := w.EksClient.DescribeFargateProfile(input)
+	if err != nil {
+		return nil, err
+	}
+	return output.FargateProfile, nil
+}
+*/
```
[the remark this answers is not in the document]
thomaswhitcomb: will do.

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,378 @@
...
+	var arn *string
+	if *instanceGroup.Spec.EKSFargateSpec.GetPodExecutionRoleArn() == "" {
+		arn, err = ctx.AwsFargateWorker.CreateDefaultRolePolicy()
+		if err != nil {
+			log.Errorf("Creation of default role policy failed: %v", err)
+			return err
+		}
+		// Save the role name of the default role we created.
+		ctx.GetInstanceGroup().Status.SetFargateRoleName(*ctx.AwsFargateWorker.RoleName)
+		log.Infof("Create() - Generated roleName: %s", *ctx.AwsFargateWorker.RoleName)
```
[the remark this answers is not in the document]
thomaswhitcomb: done

controllers/provisioners/eksfargate/eksfargate.go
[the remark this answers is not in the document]
eytan-avisror: Oh, shouldn't we update the status with the role in use regardless of whether we created it or not? I think we should always update status, and regarding the role - we should explore some mechanism such as tags to indicate ownership (not sure if its possible to tag a role), and then only delete if it's owned by controller.

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,378 @@
...
+	} else {
+		arn = instanceGroup.Spec.EKSFargateSpec.GetPodExecutionRoleArn()
+	}
+	log.Infof("Create() - Creating profile with arn: %s", *arn)
```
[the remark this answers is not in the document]
thomaswhitcomb: done

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,378 @@
...
+	return nil
+}
+func (ctx *InstanceGroupContext) Delete() error {
+	var err error
+	worker := ctx.AwsFargateWorker
+	instanceGroup := ctx.GetInstanceGroup()
+	if instanceGroup.Status.GetFargateRoleName() != "" {
+		err = worker.DeleteDefaultRolePolicy()
+		if err != nil {
+			log.Errorf("Delete() - Delete of default role policy failed: %v", err)
```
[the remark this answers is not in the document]
thomaswhitcomb: done

controllers/providers/aws/aws.go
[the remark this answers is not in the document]
eytan-avisror: Can you give an example? this call is just meant to start the creation, if we need to wait for the profile to be created, we should keep requeuing until it's ready (rather than waiting inside the reconcile)

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,378 @@
...
+		}
+	}
+	err = worker.DeleteProfile()
+	if err != nil {
+		log.Errorf("Delete() - Delete of profile failed: %v", err)
```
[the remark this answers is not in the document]
thomaswhitcomb: done

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,378 @@
...
+	updateNeeded := ctx.HasChanged()
+	instanceGroup := ctx.GetInstanceGroup()
+	if updateNeeded {
+		log.Infof("Update() - initiating a Delete due to needed updates()")
+		ctx.Delete()
+	} else {
+		instanceGroup.SetState(v1alpha1.ReconcileModified)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
thomaswhitcomb: yup. will do.

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,378 @@
...
+func (ctx *InstanceGroupContext) UpgradeNodes() error {
+	return nil
+}
+func (ctx *InstanceGroupContext) BootstrapNodes() error {
+	return nil
+}
+func (ctx *InstanceGroupContext) IsReady() bool {
+	state := ctx.AwsFargateWorker.GetState()
+	return *state.GetProfileState() == "ACTIVE"
```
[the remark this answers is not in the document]
thomaswhitcomb: yup

docs/04_instance-manager.yaml
thomaswhitcomb: whoops, didn't mean to commit that.

thomaswhitcomb: will remove that

controllers/providers/aws/aws.go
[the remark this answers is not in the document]
thomaswhitcomb: On a create, without a user-specified Role, the provisioner first creates the default Role and then creates the Profile. More often than not, the profile creation request fails because the role does not exist, even though I just created it just milliseconds before.  If I use a sleep loop, after a few trys,  the profile create request succeeds.

controllers/provisioners/eksfargate/eksfargate.go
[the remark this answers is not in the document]
thomaswhitcomb: On the request, the user specifies the Role ARN.  When the ARN is not present, I create the Role and save the Role Name in the Status.  I need the Name in order to delete the Role.  Since the user is interested in ARNs, I could save the ARN (the user-specified or generated one) in the status.  Then if the user-specified ARN is different than the ARN in the Status, then I can pull the Role name from the ARN and delete it.  Just a bit more code.

controllers/providers/aws/aws.go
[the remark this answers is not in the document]
thomaswhitcomb: I am going to pull the role creation and the profile creation apart and manage them independently through the kube polling.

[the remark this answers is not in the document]
thomaswhitcomb: I'll commit my current implementation where they are synchronized if we want it later.

thomaswhitcomb:
@eytan-avisror I pushed my support for immutable profiles.  All dependent AWS calls are handled through reconciliation. An `apply` ends up with an Error. I am going to start looking into the validating webhooks.

eytan-avisror:
@thomaswhitcomb 
Let's get this PR merged in soon (right after #83)
We can keep iterating on this before we release.
I will be making a lot of changes to shared code in the coming weeks and I don't want this PR to get too many conflicts

eytan-avisror:
I just merged #83 - there are some conflicts as a result but they are simple ones (mostly 'keep both').
Once you can resolve - we can get this merged. - validating webhooks can be a new PR

controllers/providers/aws/aws.go
```diff
@@ -507,3 +529,210 @@ func IsStackInConditionState(key string, condition string) bool {
 		return false
 	}
 }
+func IsProfileInConditionState(key *string, condition string) bool {
+	const NONE = "NONE"
```
eytan-avisror, line 533: wont it be better to simply take `key` as a string (not ptr)? since `ResourceState` is coming from AWS, all getters around such structs should probably give out copies and not a pointer. if we do `return aws.StringValue(x)` from the getter we can get values instead of pointers.

also you should not required a clause for NONE here, because `StateDiscovery` (which runs early), checks `if state.IsProvisioned()` and does the checking of whether deleted if it's false.

We should probably do work across all types to make sure we only get pointers for structs we actively modify (such as the instance group spec).

It should be OK to address this in a future PR

controllers/providers/aws/aws.go
```diff
@@ -187,6 +188,17 @@ func (w *AwsWorker) compactTags(tags []map[string]string) map[string]string {
 	return compacted
 }
 
+type AwsFargateWorker struct {
```
eytan-avisror, line 191: Can you re-use the existing AwsWorker? we should be remvoing any state from this struct (including the EKSCF ones) - state should be kept in DiscoveredState (ProfileName, ClusterName, Selectors, Subnets, Tags).
Now that we've merged #83 - consider using existing methods around IAM if they can be reused.

controllers/providers/aws/aws.go
```diff
@@ -507,3 +529,210 @@ func IsStackInConditionState(key string, condition string) bool {
...
+		FargateProfileName: w.ProfileName,
+	}
+	_, err := w.EksClient.DeleteFargateProfile(input)
+	return err
+}
+
+func (w *AwsFargateWorker) DescribeAllProfiles(profiles []*string) ([]*eks.FargateProfile, error) {
+	fargateProfiles := []*eks.FargateProfile{}
+	for _, profile := range profiles {
+		x, err := DescribeProfileWithParms(w.EksClient, w.ClusterName, profile)
```
eytan-avisror, line 698: This can get expensive from api limits perspective (imagine an account with 50 profiles that keeps reconciling them) - definitely OK for now, but we should explore caching down the road for these kind of calls.
I've created #87 to address this.

eytan-avisror: Another option is to NOT check this at all - if you only need this for the case of creation (can't create when other profile is deleting), have a requeue logic in case where the create fails due to another profile being deleted (or have a hard failure for very specific error codes).
In that case the create will fail and keep retrying (via requeue) - this way we have less code, and make less API calls.

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,262 @@
...
+			*instanceGroup.Spec.EKSFargateSpec.GetProfileName(),
+			err)
+	} else {
+		instanceGroup.SetState(v1alpha1.ReconcileModifying)
+	}
+
+	return err
+}
+func (ctx *InstanceGroupContext) CloudDiscovery() error {
+	return nil
```
eytan-avisror, line 123: You should move your AWS logic/state here.

controllers/providers/aws/aws.go
thomaswhitcomb: This is great.  I'd love to have a general rule for when to use references or values.

controllers/providers/aws/aws.go
thomaswhitcomb: So take out derivable data from AwsFargateWorker  and then merge AwsWorker and AwsFargateWorker?

eytan-avisror: Yes, but they are specific to the provisioner, hence should be part of the provisioner implementation.
I think it's better to have a single struct for AwsWorker which only holds the clients (and methods).

This needs to be done for the eks-managed provisioner as well.
We can also do this down the road if it's a big refactor

thomaswhitcomb: OK.  When I add the AwsFargateWorker methods to AwsWorker, I'll add parameters on the methods for the fields that were once in the AwsFargateWorker struct.

thomaswhitcomb: done.

controllers/providers/aws/aws.go
thomaswhitcomb: When I tried creating a profile while one was being deleted, it left the profile in some partial state. I had to delete the profile first.  So I could definitely create reconciliation logic to delete the profile when it ended up in this partial state.  I could save some state so I would know to attempt another create after deleting the profile.  WDYT?

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: will do

controllers/providers/aws/aws.go
thomaswhitcomb: Changed my mind based upon your comment below.  When I move the state logic to CloudDiscovery, I will do the NIL mapping there.

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: done

controllers/providers/aws/aws.go
thomaswhitcomb: I pulled out all the logic associated with CanDelete().  It had a race condition anyway.  I now detect the eks.ErrorCodeResourceInUse on the profile creation and deletion and rely upon the reconciliation to retry.

thomaswhitcomb:
@eytan-avisror Just added very basic create and delete for the profiles.  Approach, layout,... look ok?

test-bdd/testutil/helpers.go
```diff
@@ -35,6 +35,7 @@ import (
 type TemplateArguments struct {
 	ClusterName        string
 	KeyPairName        string
+	ProfileName        string
```
eytan-avisror, line 38: We can hardcode ProfileName in the template, is there significance in requiring to export the name as a parameter for the test? Test look good otherwise, we just need some basic assertion after create/delete that will tell us the reconcile happened as we expected, since there are no Kubernetes resources created as a result, it could be something like "And a fargate profile <name> is (found|not found) - this will also be easier by having a predictable hard-coded name for the profile.

docs/04_instance-manager.yaml
```diff
@@ -0,0 +1,734 @@
+---
```
eytan-avisror, line 1: All these docs that refer to EKSCF were completely removed in master branch
Also, there stills seems to be vendor files in this PR somehow

api/v1alpha1/instancegroup_types.go
```diff
@@ -173,6 +182,29 @@ type EKSManagedConfiguration struct {
 	Version            string              `json:"version,omitempty"`
 }
 
+// EKSCFConfiguration defines the context of an AWS Instance Group using EKSCF
+type EKSCFConfiguration struct {
```
eytan-avisror, line 186: This one was also removed in master (EKSCF deprecation)

README.md
```diff
@@ -26,14 +26,20 @@ _For installation instructions and more examples of usage, please refer to the [
...
+| eks-fargate | provision a cluster to run pods on fargate|✅|
 
 To create an instance group, submit an InstanceGroup custom resource in your cluster, and the controller will provision and bootstrap it to your cluster, and allow you to modify it from within the cluster.
 
+<<<<<<< HEAD
+### Sample Specs (by provisioner)
+#### *eks-cf* provisioner
+=======
 ### EKS sample spec (alpha-2)
+>>>>>>> master
```
eytan-avisror, line 42: Can we remove lines 37,39,40,42?

README.md
```diff
@@ -26,14 +26,20 @@ _For installation instructions and more examples of usage, please refer to the [
...
 
-| Provisioner | Description |
-| :--- | :--- |
-| eks | provision nodes on EKS |
-| eks-managed | provision managed node groups on EKS |
+| Provisioner | Description | Supported?
+| :--- | :--- | :---: |
+| eks-cf | provision nodes on EKS using cloudformation | ✅ |
+| eks-managed | provision managed node groups for EKS | ✅ |
+| eks-fargate | provision a cluster to run pods on fargate|✅|
```
eytan-avisror, line 33: Can you undo this one, and expand the existing table? I removed the "Supported?" column 😆 and modified the provisioner to reflect ekscf deprecation

test-bdd/testutil/helpers.go
thomaswhitcomb: 👍

README.md
thomaswhitcomb: yeah. whoops.

README.md
thomaswhitcomb: OK. done.

README.md
thomaswhitcomb: done

api/v1alpha1/instancegroup_types.go
thomaswhitcomb: done

docs/04_instance-manager.yaml
thomaswhitcomb: I made some local edits in docs/04_instance-manager.yaml for `image` and must have kept it out and didn't realize you removed it.  I'll remove it.

thomaswhitcomb: Are you going to make a new `instance_manager.yaml` example for the repo?

eytan-avisror: There is already one here:
https://github.com/keikoproj/instance-manager/blob/master/config/crd/bases/instance-manager-deployment.yaml

Better to keep one copy 👍

README.md
```diff
@@ -77,6 +79,103 @@ spec:
         value: some-value
 ```
 
+### EKS Fargate sample spec
```
eytan-avisror, line 82: Can we move this section to where the eks-managed spec is (314), might also be good to rename the title accordingly to `### EKS Fargate (alpha-1)`

README.md
```diff
@@ -77,6 +79,103 @@ spec:
...
+The purpose of this provisioner is to enable instance groups with AWS Fargate management.
```
eytan-avisror, line 83:
```suggestion
The purpose of this provisioner is to enable instance groups with AWS Fargate management. 
The purpose of the fargate provisioner is the enable the management of Fargate profiles.
```

README.md
```diff
@@ -77,6 +79,103 @@ spec:
...
+
+Below is an example specification for the **eks-fargate** provisioner
+
+```yaml
+apiVersion: instancemgr.keikoproj.io/v1alpha1
+kind: InstanceGroup
+metadata:
+  name: hello-world
+spec:
+  # provision for EKS using Fargate
```
eytan-avisror, line 103: Will this spec work? I think CRD mandates having spec.strategy

controllers/providers/aws/aws.go
```diff
@@ -55,7 +54,8 @@ var (
 )
 
 const (
-	IAMPolicyPrefix = "arn:aws:iam::aws:policy"
+	IAMPolicyPrefix             = "arn:aws:iam::aws:policy"
+	FargateProfileStatusMissing = "MISSING"
```
eytan-avisror, line 58: Can we not use a string for checking that it's missing? if we can do a nil check instead that would be nicer.

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+			ctx.Log.Info("Failed to get attach policy to role", "error", err)
+			return err
+		}
+		ctx.Log.Info("Attached default policy to role")
+
+	} else {
+		arn = spec.GetPodExecutionRoleArn()
+	}
+
+	ctx.Log.Info("Creating a profile.", "arn", arn)
```
eytan-avisror, line 149: can you add to all log entries the name of the instancegroup:
e.g.
```
ctx.Log.Info("Creating a profile.", "instancegroup", instanceGroup.GetName(), "arn", arn)
```
That makes debugging a bit easier when there are many

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+	tryAgain, err := ctx.AwsWorker.CreateProfile(arn)
+	if tryAgain {
+		ctx.Log.Info("Resource inuse on Create.")
```
eytan-avisror, line 152: I think we should change `tryAgain` to `profile` if possible, we can also eliminate this if condition entirely if we `return nil` at the end of this method instead of `return err`.
You can return an error inside `if err != nil` below

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+		return nil
+	}
+	if err != nil {
+		ctx.Log.Info("Creation of the fargate profile failed", "cluster",
+			spec.GetClusterName(),
+			"profile",
+			spec.GetProfileName(),
+			"error", err)
+	} else {
+		instanceGroup.SetState(v1alpha1.ReconcileModifying)
```
eytan-avisror, line 162: I think you can skip setting this - this is the state at which it enters `Create()` and if you don't modify it it will stay in that state

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+}
+func (ctx *InstanceGroupContext) Delete() error {
+	instanceGroup := ctx.GetInstanceGroup()
+	spec := instanceGroup.GetEKSFargateSpec()
+
+	worker := ctx.AwsWorker
+	if spec.GetPodExecutionRoleArn() == "" {
+		found, err := worker.DetachDefaultPolicyFromDefaultRole()
+		if found || err != nil {
+			ctx.Log.Info("Detaching the default policy", "error", err)
```
eytan-avisror, line 189: Should you log an Error here instead of Info? can you also make sure the message is more descriptive e.g. "failed to detach policy". 
Also, should we return an error if `found` is true? given that this is the delete, won't this always return true at the first pass?

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+			return err
+		}
+
+		found, err = ctx.AwsWorker.DeleteDefaultRole()
+		if found || err != nil {
+			ctx.Log.Info("Deleting the default role", "error", err)
```
eytan-avisror, line 195: Same as above

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+			return err
+		}
+	}
+	ctx.Log.Info("Deleting the profile")
+	tryAgain, err := worker.DeleteProfile()
+	if tryAgain {
```
eytan-avisror, line 201: Same comment as CreateProfile

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+	// If there is a last-applied-configuration then assume
+	// this is an update and throw an exception
+	if _, ok := annos[LastAppliedConfigurationKey]; ok {
+		return errors.New("update not supported")
+	}
+	instanceGroup.SetState(v1alpha1.ReconcileModified)
+	return nil
+}
+func (ctx *InstanceGroupContext) UpgradeNodes() error {
+	return errors.New("upgrade not supported")
```
eytan-avisror, line 232: safer to return nil here instead of an error, even though this will never be called unless you set your state to InitUpgrade somewhere

controllers/provisioners/eksfargate/eksfargate.go
```diff
@@ -0,0 +1,290 @@
...
+	return nil
+}
+func (ctx *InstanceGroupContext) IsReady() bool {
+	instanceGroup := ctx.GetInstanceGroup()
+	if instanceGroup.GetState() == v1alpha1.ReconcileModified || instanceGroup.GetState() == v1alpha1.ReconcileDeleted {
+		return true
+	}
+	return false
+}
+func (ctx *InstanceGroupContext) IsUpgradeNeeded() bool {
```
eytan-avisror, line 244: This is not a required method in the interface, you can probably remove it

controllers/provisioners/eksfargate/eksfargate_test.go
```diff
@@ -0,0 +1,952 @@
...
+	"github.com/keikoproj/instance-manager/controllers/provisioners"
+	"github.com/pkg/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	//	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	//	"k8s.io/apimachinery/pkg/runtime"
+	//	fakedynamic "k8s.io/client-go/dynamic/fake"
+	//	"k8s.io/client-go/kubernetes/fake"
```
eytan-avisror, line 36: We can probably remove these commented out lines

controllers/provisioners/eksfargate/eksfargate_test.go
```diff
@@ -0,0 +1,952 @@
...
+)
+
+type EksFargateUnitTest struct {
+	Description         string
+	Provisioner         *InstanceGroupContext
+	InstanceGroup       *v1alpha1.InstanceGroup
+	ProfileBasic        *eks.FargateProfile
+	ProfileFromDescribe *eks.FargateProfile
+	ProfileFromCreate   *eks.FargateProfile
+	//UpdateNeeded        bool
```
eytan-avisror, line 48: Same

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: "tryAgain" == true  means the CreateProfile(arn) trapped the "eks.ErrCodeResourceInUseException" exception.  This is for the case where new profiles cannot be created while others are being deleted.  So rather than bubble this up as an exception, this logic swallows the exception and handles it as a "normal" (non exception) path.  I could let the eks.ErrCodeResourceInUseException surface but it didn't seem like an error case that we would want to log and possibly alert on.  Make sense?

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: My reasoning is, if we get into UpgradeNodes, then there is definitely a bug and seems like an exception is the right response.  A nil response will hide it.

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: will do.

thomaswhitcomb: I love removing code.  Plus i get rid of a test.

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: But sure i can do it.  Just wanted you have you know my reasoning...

controllers/provisioners/eksfargate/eksfargate_test.go
thomaswhitcomb: done

controllers/provisioners/eksfargate/eksfargate_test.go
thomaswhitcomb: done

thomaswhitcomb: plus, found some others...

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: 1) found == true & no error -> implies it was successfully detached so return for a requeue.
2) found == true & error -> the detach failed for some systemic reason and we need to return the error upwards.
3) found == false & no error -> implies the detach has previously been done so proceed forward in the Delete() logic to the next resource
4) found == false & error -> this case does not exist

so the logic 
`
if found || err != nil {
                          ctx.Log.Info("Detaching the default policy", "error", err)
                          return err
                  }
`
returns immediately when found==true and the error/no error bubbles up
found == false and no error proceeds forward to get rid of the next resource.

thomaswhitcomb: I didn't want the "iam.ErrCodeNoSuchEntityException" to bubble up when calling the detach twice is the normal case.

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: Create() is entered when the state == v1alpha.ReconcileInitCreate, Right?
This logic sets the state to Modifying (from Create()) after the profile creation is successfully started.  If the profile creation fails, it stays in the InitCreate state.

thomaswhitcomb: All the creation of resources (role, policy attach and profile creation) happens in the InitCreate state, only after the last resource (in this case the profile) is created, does the state transition to ReconcileModifying...

controllers/provisioners/eksfargate/eksfargate.go
eytan-avisror: I think you should handle this logic in the provisioner, meaning the AwsWorker method should just return an error and you should handle this logic inside Create() since it is specific to your provisioner, so in AwsWorker you can send the err up as is, and in Create() you can handle the case of `ErrCodeResourceInUseException` and just return nil if it happens.

controllers/provisioners/eksfargate/eksfargate.go
eytan-avisror: I see, in this case, you should just have this SetState before you `return nil` at the end, given you make the above changes with `tryAgain`.
If all the calls inside Create() are idempotent, it's OK if it requeues while it's InitCreate, it will just retry the create.

controllers/provisioners/eksfargate/eksfargate.go
eytan-avisror: As mentioned above, we should make this logic specific to the provisioner, so you can send the error back to the caller (`Delete()`), and handle the specific case there.

This will enable you to simply return an error from `aws.go` methods and decided what to do when there is a specific error inside your provisioner

BTW, I know other methods in `aws.go` do the same thing, but we will be changing this in the future, most aws err handling should happen in the provisioner as that is where the specific logic is.

controllers/provisioners/eksfargate/eksfargate.go
eytan-avisror: Sure, I understand.
I think we prefer to 'err on the side of caution' and let the reconcile continue though, even if we accidentally add code that calls this method - otherwise the burden is on the user (he will get an error).

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: OK and you are OK if I check the exception (iam.ErrCodeNoSuchEntityException) in the Delete()?  It will make provisioners/eksfargate/eksfargate.go more dependent upon AWS.

eytan-avisror: Sure, if that's not a problem with imports, we can definitely do that,
It's also OK to make a more concrete method in `aws.go` and then handle that error there if it's more convenient. example `CreateIAMRole` would return error to caller, and `CreateIAMRoleIfNotExist` would ignore that error -> if we see multiple provisioners needing the same logic we should definitely do that.

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: ok

README.md
thomaswhitcomb: whoops.  you are right.  adding...

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: Hmmm.  Good idea, maybe later.

controllers/providers/aws/aws.go
thomaswhitcomb: I can do that but it will require passing the profile status as a pointer so I can check nil.  You OK with that?

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: will do.

controllers/provisioners/eksfargate/eksfargate.go
thomaswhitcomb: done

thomaswhitcomb: done

controllers/providers/aws/aws.go
thomaswhitcomb: will use aws.StringValue(nil) instead of FargateProfileStatusMissing

controllers/providers/aws/aws.go
```diff
@@ -700,3 +720,164 @@ func IsNodeGroupInConditionState(key string, condition string) bool {
 		return false
 	}
 }
+
+func IsStackInConditionState(key string, condition string) bool {
```
eytan-avisror, line 724: Since we removed the cloudformation provisioner, I think we can remove this one.
Are you using this anywhere?

thomaswhitcomb: no, it seems to an orphaned function.

thomaswhitcomb: removed it

test-bdd/features/01_create.feature
```diff
@@ -34,3 +34,10 @@ Feature: CRUD Create
     And the resource should converge to selector .status.currentState=ready
     And 2 nodes should be ready
 
+  @wip
```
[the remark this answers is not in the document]
thomaswhitcomb: no.  will remove the tag.

test-bdd/main_test.go
```diff
@@ -119,6 +121,7 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^the resource condition ([^"]*) should be (true|false)$`, t.theResourceConditionShouldBe)
 	s.Step(`^I (create|delete) a resource ([^"]*)$`, t.iOperateOnResource)
 	s.Step(`^I update a resource ([^"]*) with ([^"]*) set to ([^"]*)$`, t.iUpdateResourceWithField)
+	s.Step(`^the fargate profile should be (found|not found)$`, t.theProfileShouldBeFound)
```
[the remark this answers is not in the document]
thomaswhitcomb: will do

controllers/provisioners/eksfargate/types.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (ds *DiscoveredState) GetProfileStatus() string {
+	return ds.ProfileStatus
+}
+func (ds *DiscoveredState) IsProvisioned() bool {
+	return ds.GetProfileStatus() != aws.StringValue(nil)
+}
+
+type InstanceGroupContext struct {
```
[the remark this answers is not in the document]
thomaswhitcomb: yeah, i was relying on the package name and keeping it kind of polymorphic but yeah, it will make it easy to maintain if i rename it.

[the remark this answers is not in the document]
thomaswhitcomb: done

README.md
```diff
@@ -249,6 +241,126 @@ spec:
...
+            Action: 
+              - 's3:*'
+            Resource: '*'
+```
+
+AWS's Fargate Profiles are immutable.  Once one is created, it cannot be directly modified.  It first has to be deleted and then re-created with the desired change.
+
+The **eks-fargate** provisioner is built on top of that immutability.  Therefore, if an attempt is made to modify an existing profile, the provisioner will return an error.  You first have to `delete` the profile and follow that with a `create`.
+
+##### Bring your own role
```
eytan-avisror, line 343: I think this section should be under `### Features` and not under `#### EKS Fargate`

thomaswhitcomb: Reduced the title level from ##### to #### and it is under Features now.

PR #74, opened 2020-02-12
Fixed Installation Reference Walkthrough doc to have correct paths

darrendao:
At the beginning of the doc, we have the following instruction
```
cd <repository-root>  # cwd should be the root of the repository
```
The next command has the correct path for the CloudFormation template, but all of the subsequent commands have the wrong paths. This PR is to fix all the wrong paths.


darrendao:
@eytan-avisror Thanks. I've signed the CLA.

Issue #81, opened 2020-03-25
Not able to create instance group using latest release (0.5.0) and the provided sample yaml spec


darrendao:
**Is this a BUG REPORT or FEATURE REQUEST?**:
Bug

**What happened**:
Not able to create instance group using latest release (0.5.0) and the provided sample yaml spec

**What you expected to happen**:
We should be able to create instance group using the sample yaml

**How to reproduce it (as minimally and precisely as possible)**:
Using latest release (0.5.0)
Create an instance group using the instruction in the doc: https://github.com/keikoproj/instance-manager/tree/master/docs#create-an-intancegroup-object
Look at controller log and you will see the error

**Other debugging information (if applicable)**:

- controller logs:
```
time="2020-03-24T22:07:53Z" level=error msg="failed to load cloudformation configuration: template: InstanceGroup:204:38: executing \"InstanceGroup\" at <.Spec.AwsUpgradeStrategy.RollingUpgradeType.MinInstancesInService>: can't evaluate field MinInstancesInService in type *v1alpha1.RollingUpgradeStrategy"
time="2020-03-24T22:07:53Z" level=error msg="template: InstanceGroup:204:38: executing \"InstanceGroup\" at <.Spec.AwsUpgradeStrategy.RollingUpgradeType.MinInstancesInService>: can't evaluate field MinInstancesInService in type *v1alpha1.RollingUpgradeStrategy"
time="2020-03-24T22:07:53Z" level=info msg="reconcile completed with requeue"
```
I believe this is caused by this change: https://github.com/keikoproj/instance-manager/commit/5a6f76d72c345e4bfd2d71156bd7e27dfb132af5#diff-fe058806af3b4349a786152df5cbc36cR85. Since `RollingUpgradeType` is changed to a pointer, it will now be nil when we don't specify the `rollingUpdate` block in the yaml spec (rather than resolving to an empty struct with default values). I verified this by trying to create an instance group using a yaml spec that contains the `rollingUpdate` block and that worked. Example:
```
cat <<EOF | kubectl create -f -
apiVersion: instancemgr.keikoproj.io/v1alpha1
kind: InstanceGroup
metadata:
  name: hello-world
  namespace: instance-manager
spec:
  provisioner: eks-cf
  strategy:
    type: rollingUpdate
    rollingUpdate:
      maxBatchSize: 0
      minInstancesInService: 0
      minSuccessfulInstancesPercent: 0
  eks-cf:
    maxSize: 6
[...]

Issue #50, opened 2019-09-14
Pre-reconcile validation


eytan-avisror:
It would be nice if pre-reconcile (i.e. stack submission) we would attempt to validate requested objects exist and prevent a subsequent cloudformation failure (fail fast).

For example, we should verify the following exist:
      `keyPairName`
      `image`
      `clusterName` -- already being validated
      `subnets`
      `securityGroups`

Providing a wrong value in any of these will fail the stack and cause behavior described in #49

eytan-avisror:
we should use validating webhooks to block instancegroups without common required fields

PR #84, opened 2020-05-09
set RollingUpgradeStrategy if nil

RiyaJohn:
Fixing Issue #81 
by setting RollingUpgradeStrategy object so that the template gets the default values set using setRollingStrategyConfigurationDefaults()


eytan-avisror:
Thanks @RiyaJohn 
In the coming days we will fix this across all getters in a larger PR that will include other fixes such as setting default values, using validating webhooks when required fields are not provided, etc.
Ref #50 

Next release will be coming out soon and include a new provisioner which addresses many of the problems we've been seeing with the eks-cf provisioner.

PR #115, opened 2020-05-19
Pagination & Launch config deletion

### LINKED ISSUE #114: LaunchConfiguration pagination and retention
Looks like we are not paginating when we call DescribeLaunchConfigurations, we must paginate to be sure we get all launch configurations.

Also, we should do some work around the retention of launch configurations and possibly make it configurable (either globally or per CR).
Right now we are deferring a deletion but ignore failures, this can cause leak of launch configurations under some conditions

eytan-avisror:
Fixes #114 

This adds pagination for DescribeLaunchConfigurations and refactors create/update/delete to properly handle launch configuration deletion.

New logic will be called on CloudDiscovery instead of during update, and will delete the oldest launch configurations per instance group outside of the newest ones according to number of default retention (default 2) - we can make this configurable in the future.
IG Deletion now simply deletes all launch configs with the resource prefix.

- [x] BDD

### DIFF
--- a/controllers/provisioners/eks/eks.go
+++ b/controllers/provisioners/eks/eks.go
@@ -16,6 +16,7 @@ limitations under the License.
 package eks
 
 import (
+	"fmt"
 	"sync"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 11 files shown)


controllers/provisioners/eks/eks.go
```diff
@@ -26,7 +27,8 @@ import (
 )
 
 const (
-	ProvisionerName = "eks"
+	ProvisionerName                     = "eks"
+	defaultLaunchConfigurationRetention = 2
```
shrinandj, line 31: In a separate PR, we might want to make it configurable (or at least have some annotation that can override this).

eytan-avisror: Yes, I was trying to come up with a way to not add this to the CR, but using a controller flag also doesnt make much sense, because not all provisioners care about launch configurations (e.g. eks-managed).

eytan-avisror: Annotation is a good idea!

PR #125, opened 2020-05-27
Add label according to EKS cluster version

### LINKED ISSUE #120: Kubernetes 1.16: Nodes fail to join due deprecated role label
Right now the controller adds two labels by default to define the role:
- `node.kubernetes.io/role=<ROLE>` => NEW STYLE LABEL
- `node-role.kubernetes.io/<ROLE>=""` => OLD STYLE LABEL

The old style label has been deprecated in 1.16 and causes nodes to FAIL to join the cluster.
We should release a version which adds the label according to the cluster version.

- Check by cluster name what is the cluster version
- If cluster version is >= 1.16, use the new style label
- If cluster version is < 1.16, use the old style label

We will need to release a hotfix 0.6.4 with this change

garomonegro:
fixes #120


controllers/providers/aws/aws.go
```diff
@@ -407,6 +407,25 @@ func (w *AwsWorker) IsNodeGroupExist() bool {
...
+
+func (w *AwsWorker) DescribeEKSCluster(clusterName string) (*eks.Cluster, error) {
+	cluster := &eks.Cluster{} // only used in line 372 "return err, cluster", why not just directly return &eks.Cluster{}?
+	input := &eks.DescribeClusterInput{
+		Name: aws.String(clusterName),
+	}
+
+	output, err := w.EksClient.DescribeCluster(input)
+	if err != nil {
+		log.Error(err, "failed to describe cluster")
```
eytan-avisror, line 421: We can probably remove this log, if we return the error to the caller, the caller can log it if it decides to.

controllers/provisioners/eks/cloud.go
```diff
@@ -100,6 +109,16 @@ func (ctx *EksInstanceGroupContext) CloudDiscovery() error {
 		return errors.Wrap(err, "failed to describe autoscaling groups")
 	}
 
+	// Included. Temporarily marked for traceability purposes: labelbug fix @agaro
+	//######################################################################################
+	cluster, err := ctx.AwsWorker.DescribeEKSCluster(clusterName)
+	if err == nil {
```
eytan-avisror, line 115: I think this statement can be improved and get rid of the else clause, e.g.
```
cluster, err := ctx.AwsWorker.DescribeEKSCluster(clusterName)
if err != nil {
  return errors.Wrap(err, "failed to describe cluster")
}
state.SetCluster(cluster)
```

controllers/provisioners/eks/cloud.go
```diff
@@ -187,6 +206,24 @@ func (d *DiscoveredState) GetScalingGroup() *autoscaling.Group {
...
+
+// Included. Temporarily marked for traceability purposes: labelbug fix @agaro
+//######################################################################################
+func (d *DiscoveredState) SetCluster(cluster *eks.Cluster) {
+	d.Cluster = cluster
+}
+
+func (d *DiscoveredState) GetClusterVersion() (*semver.Version, error) {
+
+	ver, err := semver.NewVersion(*d.Cluster.Version)
```
eytan-avisror, line 218: You can probably use the utility `aws.StringValue` to have protection against nil pointer.
e.g.
```
version := aws.StringValue(d.Cluster.Version)
ver, err := semver.NewVersion(version)
...
```

eytan-avisror: Actually, can we also change this getter to only return `aws.StringValue(d.Cluster.Version)` and not the semver?
That way this getter does not need to return an error and we can handle the semver inside `GetLabelList`

controllers/provisioners/eks/helpers.go
```diff
@@ -156,9 +162,20 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
 	}
 	sort.Strings(labelList)
 
-	// add role label
-	for _, label := range RoleLabelsFmt {
-		labelList = append(labelList, fmt.Sprintf(label, instanceGroup.GetName()))
+	// add role label according to the cluster's k8s version
+	addOldStyleLabel := false
+	clusterVersion, err := ctx.DiscoveredState.GetClusterVersion()
```
eytan-avisror, line 167: So if we make `GetClusterVersion` not return an error as mentioned above, we can move the `semver.NewVersion` here.
Also, can we do the opposite nil check? e.g.` err != nil`, this will remove the if nesting.

controllers/provisioners/eks/helpers.go
```diff
@@ -156,9 +162,20 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
...
+	if err == nil {
+		c, err := semver.NewConstraint("< 1.16-0")
```
eytan-avisror, line 169: what does `-0` mean in this case?
I think we can rely on the EKS version being only major.minor e.g. `1.16` or `1.15`

eytan-avisror: Another possible idiom which can work here is:
```
if c, err := semver.NewConstraint("< 1.16-0"); err != nil {
  labelList = append(labelList, fmt.Sprintf(RoleNewLabelFmt, instanceGroup.GetName()))
}
```

eytan-avisror: we don't really need to handle the case of an error here, just don't append anything..
we don't really expect an error to occur here, it would mean there is no cluster version populated?

go.mod
```diff
@@ -32,4 +33,5 @@ require (
 	k8s.io/client-go v0.17.2
 	k8s.io/utils v0.0.0-20200414100711-2df71ebbae66 // indirect
 	sigs.k8s.io/controller-runtime v0.4.0
+	sigs.k8s.io/controller-tools v0.2.4 // indirect
```
eytan-avisror, line 36: Hmm, is this required?
I think you can remove this line and run `go mod tidy`

controllers/provisioners/eks/helpers.go
eytan-avisror: If we fail to `DescribeCluster`, we will get an error during CloudDiscovery.
Otherwise, we will have the cluster version - for this reason we don't need to have the getter return an error, just the value from the DiscoveredState.

In the case of error for semver.NewVersion, does it mean that the version provided is invalid or possibly empty? in this case we should log an error, and fall back to some default.
I think since the "default" behavior is adding both labels, that's what we should do, WDYT? it might mean that nodes don't join the version is actually 1.16+, but in this case, the next reconcile will remove the label.

controllers/provisioners/eks/helpers.go
eytan-avisror: Sure, that's fine to keep it in that case

go.mod
garomonegro: I did not write that, I think Visual Studio Code put it automatically.

eytan-avisror: Yeah, it's indirect - can you try removing that line and running `go mod tidy` and see if it comes back?

controllers/provisioners/eks/helpers.go
garomonegro: Sorry, I deleted the comment without noticing you had already replayed. I deleted it because you mentioned in a comment below that there is not need to handle the errors from semver, that we could just not add the "Old style role label" if there is an error.

> we don't really need to handle the case of an error here, just don't append anything..
we don't really expect an error to occur here, it would mean there is no cluster version populated?

I agree with logging the errors from semver, although I think we should just add the "New style role label" if there is an error. I think this would potentially allow nodes to join regardless the version of the cluster's k8s, while having a role label. Let me know your final thought and I will implement that.

eytan-avisror: That's fair.. 
We have two options in this case (cannot determine cluster version):
- Add both labels - if cluster is 1.16+, nodes will not join, otherwise same behavior
- Add new label only - if cluster is 1.15-, it might lose a label that was there previously, otherwise no same behavior
So the question is, what's worse - your nodes coming up with a different label (you might be relying on it in some way), or the nodes not joining at all?

In my opinion having the nodes not join will force a human to debug, otherwise, when the cluster version is discoverable in the next reconciles, it will add the correct labels and rotate the nodes so that they can join.
When we allow the nodes to join with different label we change the expected behavior. 
Keep in mind that when the nodes fail to join due to this, it will keep trying to discover the version and when it does, it will do the right thing and add the correct labels.
The end result is, we only want the nodes to join if they have the correct labels.
I'm happy to discuss further if you have a different opinion 👍

go.mod
garomonegro: It worked, thanks.

controllers/provisioners/eks/helpers.go
garomonegro: I understand now, I guess I was underestimating the importance of having the expected labels. Thank you very much.

go.mod
garomonegro: Actually, it keeps getting added. I will delete it before pushing.

controllers/provisioners/eks/eks.go
```diff
@@ -32,7 +32,9 @@ const (
 )
 
 var (
-	RoleLabelsFmt          = []string{"node.kubernetes.io/role=%s", "node-role.kubernetes.io/%s=\"\""}
+	RoleNewLabelFmt = "node.kubernetes.io/role=%s"
+	RoleOldLabelFmt = "node-role.kubernetes.io/%s=\"\""
+
```
eytan-avisror, line 37: Did we want to use two string slices here instead?
```
	RoleNewLabelFmt = []string{"node.kubernetes.io/role=%s"}
	RoleOldLabelFmt = []string{"node.kubernetes.io/role=%s", "node-role.kubernetes.io/%s=\"\""}
```

PR #129, opened 2020-05-30
Add kubernetes.io/role label via node reconciler

eytan-avisror:
This bypasses the (very annoying) mechanism put forth in https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md

This will enable the instance-group controller, to label any node that has the label `node.kubernetes.io/role=role` populated, with the respective `kubernetes.io/role=role` which cannot be added during bootstrap in 1.16.

#### 1.15
```
node-role.kubernetes.io/instance-group-1="" => added as default kubelet arguments
node.kubernetes.io/role=instance-group-1 => added as default kubelet arguments
```
#### 1.16 (nodes fail to join #120)
```
node-role.kubernetes.io/instance-group-1="" => added as default kubelet arguments
node.kubernetes.io/role=instance-group-1 => added as default kubelet arguments
```

#### 1.15
```
node-role.kubernetes.io/instance-group-1="" => added as kubelet arguments
node.kubernetes.io/instance-group-1="" => added as default kubelet arguments
```
#### 1.16
```
node.kubernetes.io/role=instance-group-1 =>  added as default kubelet arguments
```

#### 1.15
```
node.kubernetes.io/role=instance-group-1 => added as default kubelet arguments
node-role.kubernetes.io/instance-group-1="" => added as default kubelet arguments
kubernetes.io/role=instance-group-1 => added by controller
```

#### 1.16
```
node.kubernetes.io/role=instance-group-1 => added as default kubelet arguments
kubernetes.io/role=instance-group-1 => added by controller
```

#### Before control plane upgrade to 1.16
- Update controller to 0.7.0
- Nodes will now have all 3 labels
#### Upgrade control plane to 1.16
- Scale down cluster autoscaler prior to control plane upgrade, to avoid new nodes joining with the 1.15 labels
- After the upgrade, the next reconcile of the instance groups will trigger a rolling update, wait for it to complete
- Nodes will rotate to remove the old style label, you will be left with the two 1.16 labels

#### Overriding default behavior and implementing your own migration
If you want to migrate instance groups to EKS 1.16 in some different manner, you can also control the above described behavior with the following:
- By running the controller with the flag `--node-relabel` you can control whether the controller will re-label `node.kubernetes.io/role=instance-group-name` with `kubernetes.io/role=instance-group-name`, setting the flag to false will avoid this relabeling and only the default kubelet role labels will be applied, you can use this along with something such as a mutating webhook on node creation to avoid the time window introduced by the controller watch.
- If you want to override the default role labels that gets added, you can annotate your instance group with `instancemgr.keikoproj.io/default-labels: some-label/key=value,other-label/key=value`, this means the controller will not add the role labels by default, and will only add what is provided in the annotation in addition to labels provided in the spec.

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,6 +62,7 @@ func main() {
 		metricsAddr            string
 		spotRecommendationTime float64
 		enableLeaderElection   bool
+		nodeRelabel            bool
 		controllerConfPath     string
 		maxParallel            int
 		err                    error
@@ -73,6 +74,7 @@ func main() {
 	flag.StringVar(&controllerConfPath, "controller-config", "/etc/config/controller.conf", "The controller config file")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&nodeRelabel, "node-relabel", true, "relabel nodes as the join with kubernetes.io/role label via controller")
 	flag.Parse()
 	ctrl.SetLogger(zap.Logger(true))
 
@@ -119,6 +121,7 @@ func main() {
 
 	err = (&controllers.InstanceGroupReconciler{
 		SpotRecommendationTime: spotRecommendationTime,
+		NodeRelabel:            nodeRelabel,
 		Client:                 mgr.GetClient(),
 		Log:                    ctrl.Log.WithName("controllers").WithName("instancegroup"),
 		ControllerConfPath:     controllerConfPath,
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -73,6 +74,7 @@ func main() {
 	flag.StringVar(&controllerConfPath, "controller-config", "/etc/config/controller.conf", "The controller config file")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&nodeRelabel, "node-relabel", true, "relabel nodes as the join with kubernetes.io/role label via controller")
```
shrinandj, line 77: `as the` -> `as they`

Issue #119, opened 2020-05-21
Support to suspend certain autoscaling processes on ASG


SaiVishwas:
**Is this a BUG REPORT or FEATURE REQUEST?**: FEATURE REQUEST

**What happened**:
As of now, EKS and the instance-manager do not support the "[suspendProcesses](https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-suspend-resume-processes.html)" feature on an instance group the same way as Kops does as documented at https://github.com/kubernetes/kops/blob/master/docs/instance_groups.md#suspendprocess . One of the most common use cases of the "suspendProcesses" feature is to disable AZRebalance on the ASG.

**Desired state to support this feature**:
The InstanceGroup CRD contains an additional config option to accept the list of processes to be suspended.
eg. [please look for the suspendProcesses config at the bottom]
```
apiVersion: instancemgr.keikoproj.io/v1alpha1
kind: InstanceGroup
metadata:
  name: hello-world
  namespace: instance-manager
spec:
  provisioner: eks
  eks-managed:
    maxSize: 6
    minSize: 3
    configuration:
      clusterName: my-eks-cluster
      labels:
        example.label.com/label: some-value
      volSize: 20
      nodeRole: arn:aws:iam::012345678910:role/basic-eks-role
      amiType: AL2_x86_64
      instanceType: m5.large
      keyPairName: my-ec2-key-pair
      securityGroups:
      - sg-04adb6343b07c7914
      subnets:
      - subnet-0bf9bc85fd80af561
      - subnet-0130025d2673de5e4
      - subnet-01a5c28e074c46580
      suspendProcesses:
      - AZRebalance
      tags:
      - key: my-ec2-tag
        value: some-value
```

Hence the need for the "suspendProcesses" config option in the InstanceGroup CRD, which is then picked up by the instance-manager controller. Since, in EKS an instance group maps to an ASG, the controller then updates the ASG accordingly.

**How would you implement this feature**:

[...]

PR #132, opened 2020-06-02
Suspend processes

SaiVishwas:
Fixes https://github.com/keikoproj/instance-manager/issues/119

Assumption: In the desired/target state, only the processes mentioned in the input yaml need to be suspended on the ASG
**Scenarios tested:**
**During Create**
1. When the suspendProcesses config is not present, then no action taken
2. When the config is passed, post ASG creation, the processes mentioned are suspended using a single AWS API call

**During Update**

1. When the target list mentioned in yaml is different from current list of suspended processes in ASG, only then it is updated (resumed/suspended)

@eytan-avisror please review and let me know if you would like me to improve anything.
Thank you


api/v1alpha1/instancegroup_types.go
```diff
@@ -153,6 +153,7 @@ type EKSConfiguration struct {
 	NodeSecurityGroups          []string            `json:"securityGroups,omitempty"`
 	Volumes                     []NodeVolume        `json:"volumes,omitempty"`
 	Subnets                     []string            `json:"subnets"`
+	SuspendedProcesses          []string            `json:"suspendProcesses"`
```
eytan-avisror, line 156: Can you please add omitempty to this field

api/v1alpha1/instancegroup_types.go
```diff
@@ -372,6 +373,12 @@ func (c *EKSConfiguration) GetSubnets() []string {
 	}
 	return c.Subnets
 }
+func (c *EKSConfiguration) GetSuspendProcesses() []string {
+	if c.SuspendedProcesses == nil {
```
eytan-avisror, line 377: I don't think this can be nil, just an empty slice

controllers/providers/aws/aws.go
```diff
@@ -201,6 +201,44 @@ func (w *AwsWorker) DeleteScalingGroup(name string) error {
...
 
+func (w *AwsWorker) SetSuspendProcesses(name string, processesToSuspend []string) error {
+
+	var scalingProcesses []*string
+
+	for _, process := range processesToSuspend {
+		scalingProcesses = append(scalingProcesses, aws.String(process))
+	}
+
+	input := &autoscaling.ScalingProcessQuery{
```
eytan-avisror, line 212: Can you move this call out to a wrapper method under `aws.go`?

controllers/providers/aws/aws.go
```diff
@@ -201,6 +201,44 @@ func (w *AwsWorker) DeleteScalingGroup(name string) error {
...
+
+func (w *AwsWorker) SetResumeProcesses(name string, processesToResume []string) error {
+
+	var scalingProcesses []*string
+
+	for _, process := range processesToResume {
+		scalingProcesses = append(scalingProcesses, aws.String(process))
+	}
+
+	input := &autoscaling.ScalingProcessQuery{
```
eytan-avisror, line 231: Same

controllers/provisioners/eks/create.go
```diff
@@ -89,6 +89,14 @@ func (ctx *EksInstanceGroupContext) CreateScalingGroup(lcName string) error {
 	}
 	ctx.Log.Info("created scaling group", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName)
 
+	if spec.EKSConfiguration.SuspendedProcesses != nil {
```
eytan-avisror, line 92: I think you can probably remove this if, and add a length check inside the method/s
```
if len(processesToSuspend) == 0 {
  return nil
}
```
inside `SetSuspendProcesses` - same for `SetResumeProcesses`

controllers/provisioners/eks/update.go
```diff
@@ -116,6 +116,27 @@ func (ctx *EksInstanceGroupContext) UpdateScalingGroup() error {
 		ctx.Log.Info("updated scaling group tags", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName)
 	}
 
+	isSuspendProcessesUpdateNeeded, processesToSuspend, processesToResume := ctx.SuspendProcessesUpdateNeeded()
```
eytan-avisror, line 119: Can this be just a boolean predicate instead?
You can move all the logic of finding which processes to suspend or resume to a method inside `helpers.go` e.g. `UpdateScalingProcesses`.
So that Create/Update calls > UpdateScalingProcesses and it calls > AwsWorker.SetSuspendProcesses or AwsWorker.SetResumeProcesses according to whether there is a diff
Example here:
https://github.com/keikoproj/instance-manager/blob/master/controllers/provisioners/eks/helpers.go#L428

controllers/provisioners/eks/update.go
```diff
@@ -206,6 +227,43 @@ func (ctx *EksInstanceGroupContext) ScalingGroupUpdateNeeded() bool {
...
+	for _, item := range b {
+		m[item] = true
+	}
+
+	for _, item := range a {
+		if _, ok := m[item]; !ok {
+			diff = append(diff, item)
+		}
+	}
+	return
```
eytan-avisror, line 264: should this return a diff?

Can you also move this method to common package?

controllers/providers/aws/aws.go
```diff
@@ -201,6 +201,44 @@ func (w *AwsWorker) DeleteScalingGroup(name string) error {
...
+		AutoScalingGroupName: aws.String(name),
+		ScalingProcesses:     scalingProcesses,
```
eytan-avisror, line 233: You can probably just use `aws.StringSlice(processesToResume)` and remove the iteration in the top

controllers/providers/aws/aws.go
```diff
@@ -201,6 +201,44 @@ func (w *AwsWorker) DeleteScalingGroup(name string) error {
...
+
+	var scalingProcesses []*string
+
+	for _, process := range processesToSuspend {
+		scalingProcesses = append(scalingProcesses, aws.String(process))
+	}
+
+	input := &autoscaling.ScalingProcessQuery{
+		AutoScalingGroupName: aws.String(name),
+		ScalingProcesses:     scalingProcesses,
```
eytan-avisror, line 214: You can probably just use `aws.StringSlice(processesToSuspend)` and remove the iteration in the top

eytan-avisror:
Thanks for the PR and contribution @SaiVishwas 
Some minor comments added.

api/v1alpha1/instancegroup_types.go
SaiVishwas: Hi @eytan-avisror ,
While debugging, I see that it is nil and has hence reached line no. 378

Here's a snapshot of its value when the input yaml contains no mention of the "suspendProcesses" field

Screen Shot 2020-06-03 at 11 41 46 AM

Please do let me know if I should leave this piece as is or whether you would like me to perform any modifications to it.

Thank you.

SaiVishwas: Hi @eytan-avisror ,
Would you prefer the 
`if len(c.SuspendedProcesses) == 0:` instead of `c.SuspendedProcesses == nil` ? 

I see that checking length of a nil slice doesn't result in an error in golang (https://tour.golang.org/moretypes/12) 

Please let me know which one you prefer.

Thank you

controllers/provisioners/eks/create.go
```diff
@@ -89,6 +89,14 @@ func (ctx *EksInstanceGroupContext) CreateScalingGroup(lcName string) error {
 	}
 	ctx.Log.Info("created scaling group", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName)
 
+	if len(spec.EKSConfiguration.SuspendedProcesses) > 0 {
```
eytan-avisror, line 92: We already do a len check in the actual methods, is it possible to just skip this if and always call UpdateScalingProcesses?
Minor nit: maybe nicer to call it as:
```
if err := ctx. UpdateScalingProcesses(asgName); err != nil {
		return err
}
```
you can probably move the logging to inside `UpdateScalingProcesses` and also specify what changed.
e.g.
```
ctx.Log.Info("suspended scaling processes", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName, "difference", diff)
---
ctx.Log.Info("resumed scaling processes", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName, "difference", diff)
```

controllers/provisioners/eks/update.go
SaiVishwas: Hi,
I have moved it to the common package. I had previously used a [naked return](https://www.ardanlabs.com/blog/2013/10/functions-and-naked-returns-in-go.html)
Have refactored it to use the conventional return style.

Thank you.

controllers/provisioners/eks/create.go
SaiVishwas: Done 👍

api/v1alpha1/instancegroup_types.go
```diff
@@ -153,6 +153,7 @@ type EKSConfiguration struct {
 	NodeSecurityGroups          []string            `json:"securityGroups,omitempty"`
 	Volumes                     []NodeVolume        `json:"volumes,omitempty"`
 	Subnets                     []string            `json:"subnets"`
+	SuspendedProcesses          []string            `json:"suspendProcesses,omitempty"`
```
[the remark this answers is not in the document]
SaiVishwas: Done 👍

SaiVishwas:
Hi @eytan-avisror ,
I have made the suggested changes. The newly added methods are being tested as part of the unit tests. I went through the code coverage report and drop in coverage is due to the 
```
err != nil { return err }
```
code block.

Thank you

controllers/provisioners/eks/update.go
```diff
@@ -116,6 +116,12 @@ func (ctx *EksInstanceGroupContext) UpdateScalingGroup() error {
 		ctx.Log.Info("updated scaling group tags", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName)
 	}
 
+	if ctx.SuspendProcessesUpdateNeeded() {
```
eytan-avisror, line 119: One more question.. 
Since inside of `SetSuspendProcesses` or `SetResumeProcesses` we use:
```
if len(processesToSuspend) == 0 {
		return nil
	}
```
Does that mean we can get rid of the predicate and allow `UpdateScalingProcesses` to always be called?

eytan-avisror:
Thanks @SaiVishwas 
We should be able to merge this shortly, had one more question above.

controllers/provisioners/eks/update.go
SaiVishwas: Hi @eytan-avisror ,
Yes, that's right. Post the code refactor this check is no longer required. I have removed the check and corresponding method.

Thank you.

controllers/provisioners/eks/helpers.go
```diff
@@ -110,6 +110,37 @@ func (ctx *EksInstanceGroupContext) GetRemovedTags(asgName string) []*autoscalin
...
+	for _, element := range scalingGroup.SuspendedProcesses {
+		groupSuspendProcesses = append(groupSuspendProcesses, *element.ProcessName)
+	}
+
+	processesToSuspend := common.Difference(specSuspendProcesses, groupSuspendProcesses)
+	err := ctx.AwsWorker.SetSuspendProcesses(asgName, processesToSuspend)
+	if err != nil {
+		return err
+	}
+	ctx.Log.Info("suspended scaling processes", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName, "difference", processesToSuspend)
```
eytan-avisror, line 132: last thing! 💯 

Can we change it so it only logs when a change actually happens?
for example:
```
if suspend := common.Difference(specSuspendProcesses, groupSuspendProcesses); len(suspend) > 0 {
  if err := ctx.AwsWorker.SetSuspendProcesses(asgName, processesToSuspend); err != nil {
    return err
  }
  ctx.Log.Info("suspended scaling processes", "instancegroup", instanceGroup.GetName(), "scalinggroup", asgName, "processes", suspend)
}
```
same for resume - this would mean we can actually remove the len check from `AwsWorker.SetSuspendProcesses` into here.
also, can we change the key "difference" in the log to "processes" - it might be more accurate.

Sorry for the long review process!

SaiVishwas: Hi @eytan-avisror ,
Absolutely no problem, please do let me know if there are any other improvements that you would be interested in.

I have made the above change, so it logs only when there is a change. The len checks in AwsWorker.SetSuspendProcesses and AwsWorker.SetResumeProcesses have been moved to the helper.

Thank you.

eytan-avisror:
15 scenarios (15 passed)
72 steps (72 passed)
22m41.99080058s
PASS
ok  	github.com/keikoproj/instance-manager/test-bdd	1362.315s [no tests to run]

PR #143, opened 2020-07-01
Added sleep after attach and ensured profile name does not start with…

thomaswhitcomb:
… 'eks'

### DIFF
--- a/config/crd/bases/instancemgr.keikoproj.io_instancegroups.yaml
+++ b/config/crd/bases/instancemgr.keikoproj.io_instancegroups.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.9
+    controller-gen.kubebuilder.io/version: v0.2.4
   creationTimestamp: null
   name: instancegroups.instancemgr.keikoproj.io
 spec:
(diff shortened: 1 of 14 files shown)


config/crd/bases/instancemgr.keikoproj.io_instancegroups.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.9
+    controller-gen.kubebuilder.io/version: v0.2.4
```
eytan-avisror, line 7: Looks like you need to run `make controller-gen` followed by a `make manifests` - you are using an older version of controller-gen

go.mod
```diff
@@ -32,4 +32,5 @@ require (
 	k8s.io/client-go v0.17.2
 	k8s.io/utils v0.0.0-20200414100711-2df71ebbae66 // indirect
 	sigs.k8s.io/controller-runtime v0.4.0
+	sigs.k8s.io/controller-tools v0.2.9 // indirect
```
eytan-avisror, line 35: Mmm.. not sure where this comes from? we need controller-gen 0.2.9, not controller-tools.
Also it's installed locally and not as a dependency.
Does it stay if you run `go mod tidy`?

thomaswhitcomb: It goes away after running `go mod tidy`

thomaswhitcomb: commit and push it?

eytan-avisror: yep

PR #146, opened 2020-07-08
User data support

### LINKED ISSUE #104: Add UserData support for EKS
Feature Request

**What happened**:

Can't find where to place the UserData to inject into the Autoscale Group through Cloudformation.

Be able to inject UserData to CloudFormation

- Kubernetes version:

1.15 , 1.16
```

eytan-avisror:
Fixes #104 

TBD
- [x] BDD passing 
- [x] Documentation
- [x] Unit tests


kevdowney:
Does this also solve for someone who wants to create a `userData` for systemd targets, could be for very early targets like `local-fs-pre.target`?
See https://www.freedesktop.org/software/systemd/man/bootup.html#System%20Manager%20Bootup

eytan-avisror:
> Does this also solve for someone who wants to create a `userData` for systemd targets, could be for very early targets like `local-fs-pre.target`?
> See https://www.freedesktop.org/software/systemd/man/bootup.html#System%20Manager%20Bootup

UserData is a LaunchConfiguration element, this just let's use inject a script that runs when the node boots up, depending on a stage, before/after it's bootstrapped to a cluster.

You can run whatever bash code you want in these stages

PR #159, opened 2020-08-06
Resolve security groups / subnets by Name tag

### LINKED ISSUE #156: EKS: Take resource names in addition to IDs
We should accept resource names according to Name tag in addition to resource ID for the following fields:
- securityGroups
- subnets

eytan-avisror:
Fixes #156 

This adds support for using "Name" tag of resource instead of ID for Subnets & Security Groups if provided.

- [x] Functional test passing

### DIFF
--- a/controllers/provisioners/eks/update.go
+++ b/controllers/provisioners/eks/update.go
@@ -84,7 +84,6 @@ func (ctx *EksInstanceGroupContext) UpdateScalingGroup() error {
 	var (
 		instanceGroup = ctx.GetInstanceGroup()
 		spec          = instanceGroup.GetEKSSpec()
-		configuration = instanceGroup.GetEKSConfiguration()
 		status        = instanceGroup.GetStatus()
 		state         = ctx.GetDiscoveredState()
 		scalingGroup  = state.GetScalingGroup()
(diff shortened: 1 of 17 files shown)


controllers/provisioners/eks/update.go
```diff
@@ -99,7 +98,7 @@ func (ctx *EksInstanceGroupContext) UpdateScalingGroup() error {
 			LaunchConfigurationName: aws.String(state.GetActiveLaunchConfigurationName()),
 			MinSize:                 aws.Int64(spec.GetMinSize()),
 			MaxSize:                 aws.Int64(spec.GetMaxSize()),
-			VPCZoneIdentifier:       aws.String(common.ConcatenateList(configuration.GetSubnets(), ",")),
+			VPCZoneIdentifier:       aws.String(common.ConcatenateList(ctx.ResolveSubnets(), ",")),
```
kevdowney, line 101: When this list is empty, when errors occur, will this call fail?

eytan-avisror: Yes, this field is required, and there is validation for it, i.e. it will fail much earlier if no subnets are provided.

PR #161, opened 2020-08-07
Additional volume options

### LINKED ISSUE #155: Allow IOPS config option for io1 volumeType
FEATURE
**What happened**:
Allow IOPS config option for io1 volumeType 

https://github.com/keikoproj/instance-manager/blob/5843c10e7b365f7206542e08efd754413d51799c/api/v1alpha1/instancegroup_types.go#L193-L196

More Details about io1
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html

**Environment**:

- Kubernetes version:

```bash
kubectl version -o yaml
```

- InstanceGroup status:

```bash
```

- controller logs:

```bash
```

eytan-avisror:
Fixes #155 

Adds additional volume options to NodeVolumes


controllers/providers/aws/aws.go
```diff
@@ -151,15 +153,32 @@ func (w *AwsWorker) InstanceProfileExist(name string) (*iam.InstanceProfile, boo
...
+	if delete != nil {
+		device.Ebs.DeleteOnTermination = delete
+	} else {
+		device.Ebs.DeleteOnTermination = aws.Bool(true)
+	}
+	if encrypt != nil {
+		device.Ebs.Encrypted = encrypt
+	}
+	if iops != 0 {
+		device.Ebs.Iops = aws.Int64(iops)
```
vgunapati, line 172: can we add this validation https://github.com/aws/aws-sdk-go/blob/v1.33.16/service/autoscaling/api.go#L9756

eytan-avisror: I can add a validation that IOPS is not 1-99

eytan-avisror:
> This feature is to gather IOPS, DeleteonTermination and Encrypted info from AWS right ?

It's to allow these fields to be configurable in the custom resource

controllers/providers/aws/aws.go
eytan-avisror: added

shaoxt:
OK, what if the old configuration doesn't have UserDataStage/data or UserDataStage/stage ?
Will it causes failure or the validation could take care of the case ?

> > This feature is to gather IOPS, DeleteonTermination and Encrypted info from AWS right ?
> 
> It's to allow these fields to be configurable in the custom resource

eytan-avisror:
> OK, what if the old configuration doesn't have UserDataStage/data or UserDataStage/stage ?
> Will it causes failure or the validation could take care of the case ?
> 
> > > This feature is to gather IOPS, DeleteonTermination and Encrypted info from AWS right ?
> > 
> > 
> > It's to allow these fields to be configurable in the custom resource

The UserDataStage type changes are backwards compatible, we only added a "name" field and this PR also makes it 'omitempty' to make sure if an old CR did not have this field it won't break.
The name field we added is for convenience any way and is not mandatory

PR #169, opened 2020-09-08
Add mountOptions to NodeVolume to initialize volumes

### LINKED ISSUE #167: Add mountOptions to volumes
We should add an `mountOptions` to volumes to support use cases of initializing a file system and mounting it.
The root volume is already initialized by the AMI, but additional EBS volumes only gets attached to the EC2 - which is pretty useless for the user.

Current API
```
volumes:
  - name: <string> : represents the device name, e.g. /dev/xvda (required)
  type: <string> : represents the type of volume, must be one of supported types "standard", "io1", "gp2", "st1", "sc1" (required)
  size: <int64> : represents a volume size in gigabytes, cannot be used with snapshotId
  snapshotId : <string> : represents a snapshot ID to use, cannot be used with size
  iops: <int64> : represents number of IOPS to provision volume with (min 100)
  deleteOnTermination : <bool> : delete the EBS volume when the instance is terminated (defaults to true)
  encrypted: <bool> : encrypt the EBS volume with a KMS key
```

Proposed changes
```
volumes:
  - name: /dev/xvdb
  type: gp2
  size: 32
  iops: 100
  deleteOnTermination: true
  encrypted: true

  # describes options for automatic formatting/mounting
  mountOptions:
    fileSystem: XFS    # support XFS or EXT4
    mount: /data         # where to mount the volume
    persistance: true  # should volume persist after reboots
```

using mountOptions will allow a user to customize a specific EBS volume and make it usable.
This will inject pre-bootstrap userdata to make sure the volume is formatted and mounted by using `mkfs` and modify `/etc/fstab` for persistence

Reference:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-using-volumes.html

eytan-avisror:
Fixes #167 

This PR adds support for mountOptions under NodeVolume struct, which when used injects commands into UserData to create a filesystem and mount it.

TBD:
- [x] Unit test
- [x] Docs
- [x] BDD Passing


vgunapati (approved):
LGTM

PR #183, opened 2020-11-02
Add capability to automatically add cluster autoscaler ASG tags

### LINKED ISSUE #180: Feature Request - native Cluster Autoscaler support
Feature Request
**What happened**:
It would be nice if cluster autoscaler was supported natively. 
Today, registering instance groups with cluster autoscaler can be achieved through the manual addition of Tags in the CR, but this could be cumbersome for users to add to each and every CR - especially since most of the information is already present in the CR. 

**Proposal**
Add a new field like "useClusterAutoscaler" to the InstanceGroup defintion which, if set, would automatically tag the ASG with the following tags. 

```
k8s.io/cluster-autoscaler/<cluster-name>: owned
k8s.io/cluster-autoscaler/enabled: true
k8s.io/cluster-autoscaler/node-template/label/<my-label-key>: <my-label-value> (this would be repeated for each unique label) 
k8s.io/cluster-autoscaler/node-template/taint/my-taint: "true:NoSchedule" (this would be repeated for each unique taint)
```

backjo:
Closes #180

---
This change is [Reviewable](https://reviewable.io/reviews/keikoproj/instance-manager/183)

### DIFF
--- a/controllers/provisioners/eks/helpers.go
+++ b/controllers/provisioners/eks/helpers.go
@@ -206,6 +206,9 @@ func (ctx *EksInstanceGroupContext) GetAddedTags(asgName string) []*autoscaling.
 		instanceGroup = ctx.GetInstanceGroup()
 		configuration = instanceGroup.GetEKSConfiguration()
 		clusterName   = configuration.GetClusterName()
+		annotations   = instanceGroup.GetAnnotations()
+		labels 	      = configuration.GetLabels()
+		taints        = configuration.GetTaints()
 	)
 
 	tags = append(tags, ctx.AwsWorker.NewTag("Name", asgName, asgName))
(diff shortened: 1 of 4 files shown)


controllers/provisioners/eks/helpers.go
```diff
@@ -214,6 +214,21 @@ func (ctx *EksInstanceGroupContext) GetAddedTags(asgName string) []*autoscaling.
...
+			tags = append(tags, ctx.AwsWorker.NewTag("k8s.io/cluster-autoscaler/node-template/label/" + label, labelValue, asgName))
+		}
+
+		for _, taint := range instanceGroup.Spec.EKSSpec.EKSConfiguration.Taints {
+			tagValue := taint.Value + ":" + string(taint.Effect)
+			tag := ctx.AwsWorker.NewTag("k8s.io/cluster-autoscaler/node-template/taint/" + taint.Key, tagValue, asgName)
+			tags = append(tags, tag)
+		}
+	}
+
```
eytan-avisror, line 231: PR looks great! just some (very) minor change suggestions just to follow existing code style:

- Can we add `annotations   = instanceGroup.GetAnnotations()` at the top level var declaration and use `annotations[ClusterAutoscalerEnabledAnnotation]` instead of `instanceGroup.ObjectMeta.Annotations`

- Same for `instanceGroup.Spec.EKSSpec.EKSConfiguration.Labels`, you can use `configuration.GetLabels()` instead

- Same for `instanceGroup.Spec.EKSSpec.EKSConfiguration.Taints`, you can use `configuration.GetTaints()`

- Consider using `fmt.Sprintf` instead of string concatenation - it might be a bit slower, but perhaps makes code bit more readable and also removes the need to do `string(taint.Effect)`? e.g. 
```go
tagValue := taint.Value + ":" + string(taint.Effect)
``` 
vs 
```go
tagValue := fmt.Sprintf("%v:%v", taint.Value, taint.Effect)
```

controllers/provisioners/eks/helpers_test.go
```diff
@@ -31,6 +32,54 @@ import (
...
+		k       = MockKubernetesClientSet()
+		ig      = MockInstanceGroup()
+		asgMock = NewAutoScalingMocker()
+		iamMock = NewIamMocker()
+		eksMock = NewEksMocker()
+		ec2Mock = NewEc2Mocker()
+	)
+
+	w := MockAwsWorker(asgMock, iamMock, eksMock, ec2Mock)
+
```
eytan-avisror, line 46: You can probably use same getters suggested above in your tests, e.g. ig.GetAnnotations, configuration.GetLabels, etc

eytan-avisror (approved):
LG

PR #186, opened 2020-11-05
Initial basic support for osFamily label and Windows nodes

backjo:
This PR adds support for a new label - `instancemgr.keikoproj.io/os-family` - which indicates that the targeted os needs to be handled differently than how the default behavior. This label is initially being used to differentiate between bootstrap and aws-auth behavior for Windows nodes, but could also be used for managing bootstrap/aws-auth differences for BottleRocket or other image types in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/keikoproj/instance-manager/186)


controllers/common/bootstrap.go
```diff
@@ -22,59 +22,68 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func GetNodeBootstrapUpsert(arn string) *awsauth.MapperArguments {
+func GetGroupsForOsFamily(osFamily string) []string {
+	if osFamily == "windows" {
```
eytan-avisror, line 26: add case safety
```suggestion
	if strings.EqualFold(osFamily, "windows") {
```

controllers/providers/kubernetes/utils.go
```diff
@@ -254,6 +254,11 @@ func ParseCustomResourceYaml(raw string) (*unstructured.Unstructured, error) {
 
 func ConfigmapHash(cm *corev1.ConfigMap) string {
 	var buf strings.Builder
+	defer func() {
```
eytan-avisror, line 257: Did you see this function getting a panic?

backjo: yeah, meant to not include this - it was causing panics running in GoLand debug mode - but not outside of that.

controllers/provisioners/eks/helpers.go
```diff
@@ -97,8 +97,19 @@ func (ctx *EksInstanceGroupContext) ResolveSecurityGroups() []string {
 }
 
 func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, payload UserDataPayload, mounts []MountOpts) string {
-
-	var UserDataTemplate = `#!/bin/bash
+	osFamily := ctx.GetInstanceGroup().Labels[OsFamilyLabel]
+	var UserDataTemplate = ""
+	if osFamily == "windows" {
```
eytan-avisror, line 102:
```suggestion
	if strings.EqualFold(osFamily, "windows") {
```

controllers/provisioners/eks/helpers.go
```diff
@@ -379,7 +395,7 @@ func (ctx *EksInstanceGroupContext) GetBootstrapArgs() string {
 
 	labelsFlag := fmt.Sprintf("--node-labels=%v", strings.Join(ctx.GetLabelList(), ","))
 	taintsFlag := fmt.Sprintf("--register-with-taints=%v", strings.Join(ctx.GetTaintList(), ","))
-	return fmt.Sprintf("--kubelet-extra-args '%v %v %v'", labelsFlag, taintsFlag, bootstrapArgs)
+	return fmt.Sprintf("%v %v %v", labelsFlag, taintsFlag, bootstrapArgs)
```
eytan-avisror, line 398: Are we changing existing behavior here?

eytan-avisror: NVM, saw you wrapped it in 
```
func (ctx *EksInstanceGroupContext) GetBootstrapArgs() string {
	return fmt.Sprintf("--kubelet-extra-args '%v'", ctx.GetKubeletExtraArgs())
}
```

backjo: Nope - this just seperates the building of the kubelet arg string from the placement of that into the bootstrap script.

controllers/provisioners/eks/eks.go
```diff
@@ -31,6 +31,7 @@ const (
 	ProvisionerName                     = "eks"
 	defaultLaunchConfigurationRetention = 2
 	OverrideDefaultLabelsAnnotationKey  = "instancemgr.keikoproj.io/default-labels"
+	OsFamilyLabel				        = "instancemgr.keikoproj.io/os-family"
```
eytan-avisror, line 34: WDYT regarding label vs. annotation.
Would it be confusing that some configurations are via annotations and others via labels?

controllers/provisioners/eks/eks.go
```diff
@@ -108,6 +110,19 @@ func (ctx *EksInstanceGroupContext) GetInstanceGroup() *v1alpha1.InstanceGroup {
 	return &v1alpha1.InstanceGroup{}
 }
 
+func (ctx *EksInstanceGroupContext) GetOsFamily() string {
+	if ctx != nil && ctx.InstanceGroup != nil {
```
eytan-avisror, line 114: We probably don't need a nil check here because you cannot get to this code path without an InstanceGroup or context object being populated.

Very minor, but might be a bit nicer to use 
```golang
var (
  instanceGroup  = ctx.GetInstanceGroup()
  labels = instanceGroup.GetLabels()
)
```
And also have more safety around getting the label:

```golang
if ok, _ := labels[OsFamilyLabel]; ok {
  return labels[OsFamilyLabel]
} 
return "default"  // return "" is will also be fine here
```

controllers/provisioners/eks/helpers.go
```diff
@@ -97,8 +97,19 @@ func (ctx *EksInstanceGroupContext) ResolveSecurityGroups() []string {
 }
 
 func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, payload UserDataPayload, mounts []MountOpts) string {
-
-	var UserDataTemplate = `#!/bin/bash
+	osFamily := ctx.GetInstanceGroup().Labels[OsFamilyLabel]
+	var UserDataTemplate = ""
```
eytan-avisror, line 101:
```suggestion
	var UserDataTemplate string
```

controllers/provisioners/eks/eks.go
backjo: Yeah, i'm on the fence on this one as well. 

Arguments for an annotation:
- Aligns with how similar configurations are enabled. 

Arguments for a label:
- Selectable. We could, in the future, allow cluster operators to define 'defaults' in the ConfigMap based off of the family type (or any other selector). So I, as an operator, could define that for os-family=windows, use the windows AMI, windows SG etc - and namespace operators would just need to label their IG with the appropriate os-type.

controllers/provisioners/eks/helpers.go
```diff
@@ -97,8 +97,19 @@ func (ctx *EksInstanceGroupContext) ResolveSecurityGroups() []string {
 }
 
 func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, payload UserDataPayload, mounts []MountOpts) string {
-
-	var UserDataTemplate = `#!/bin/bash
+	osFamily := ctx.GetInstanceGroup().Labels[OsFamilyLabel]
```
eytan-avisror, line 100: This should work right?
```suggestion
      osFamily := ctx.GetOsFamily()
```

controllers/provisioners/eks/eks.go
eytan-avisror: Do you need a selector for IGs? I would think you want to add this as a node label no?

backjo: You do for the mentioned cluster operator use case, I think. 

Some way to group/filter IGs to apply default settings - i.e. windows node IGs get image: x, default node IGs get image y, bottlerocket node IGs get image z.

eytan-avisror: Yeah, I guess we haven't decided / designed how the second part of AMI selection will work and whether it will use labelSelector to do that - the way the default /configmap feature is designed right now is not via updating / patching CRs, so if you are referring to the existing feature with the configmap the labelSelector won't help here.
I think we need to spend more time thinking how the AMI selection will look like

If we don't need it to be selectable, will annotation be more appropriate? if not, we can stick with a label.

backjo: Yep, that sounds correct to me.

backjo: I'll put it as an annotation for now, actually - to keep this independent of the conversation around ami selection / multiple defaults.

eytan-avisror:
@backjo LG 🥇 
Can you just also add some unit tests? you can use `make coverage` to generate a coverage report

eytan-avisror:
@backjo This seems mergeable to me.
I think we only need to remove the defer panic unless we think it's needed and also remove the empty file `test-bdd/templates/instance-group-windows.yaml` unless you want to add a functional test as well? This might be painful because as of now in order to run the test you need a cluster with both upgrade-manager & minion-manager. If you want we can add the BDD later - I intend to make the functional test automated per PR once we move to Github Actions

backjo:
@eytan-avisror sounds good - I went ahead and removed the BDD. I've got a cluster with most of the right stuff on it, which is how I've been testing, but it needs a bit more work.

eytan-avisror (approved):
Great stuff

PR #188, opened 2020-11-09
feat: Add BottleRocket os family

backjo:
This PR leverages the os-family annotation introduced in #186 to support [BottleRocket](https://aws.amazon.com/bottlerocket/) AMIs on EKS self-managed node groups. 

### DIFF
--- a/controllers/provisioners/eks/create_test.go
+++ b/controllers/provisioners/eks/create_test.go
@@ -46,7 +46,13 @@ func TestCreateManagedRolePositive(t *testing.T) {
 	w := MockAwsWorker(asgMock, iamMock, eksMock, ec2Mock)
 	ctx := MockContext(ig, k, w)
 	state := ctx.GetDiscoveredState()
-	state.SetCluster(&eks.Cluster{Version: aws.String("1.15")})
+	state.SetCluster(&eks.Cluster{
+		Version: aws.String("1.15"),
+		CertificateAuthority: &eks.Certificate{
+			Data: aws.String("clusterca"),
+		},
+		Endpoint: aws.String("foo.amazonaws.com"),
+	})
 	state.Publisher.Client = k.Kubernetes
 	state.ScalingConfiguration = &scaling.LaunchConfiguration{
 		AwsWorker: w,
@@ -136,6 +142,12 @@ func TestCreateScalingGroupPositive(t *testing.T) {
 		Publisher: kubeprovider.EventPublisher{
 			Client: k.Kubernetes,
 		},
+		Cluster: &eks.Cluster{
+			CertificateAuthority: &eks.Certificate{
+				Data: aws.String("foo"),
+			},
+			Endpoint: aws.String("foo.amazonaws.com"),
+		},
 		ScalingConfiguration: lc,
 	})
 
@@ -175,6 +187,13 @@ func TestCreateNoOp(t *testing.T) {
 	}
 
 	ctx.SetDiscoveredState(&DiscoveredState{
+		Cluster: &eks.Cluster{
+			CertificateAuthority: &eks.Certificate{
+				Data: aws.String(""),
+			},
+			Endpoint:           aws.String("foo.amazonaws.com"),
+			ResourcesVpcConfig: &eks.VpcConfigResponse{},
+		},
 		Publisher: kubeprovider.EventPublisher{
 			Client: k.Kubernetes,
 		},
--- a/controllers/provisioners/eks/create_test.go
+++ b/controllers/provisioners/eks/create_test.go
@@ -25,7 +25,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
-	"github.com/aws/aws-sdk-go/service/eks"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/keikoproj/instance-manager/api/v1alpha1"
 	"github.com/onsi/gomega"
@@ -46,13 +45,7 @@ func TestCreateManagedRolePositive(t *testing.T) {
 	w := MockAwsWorker(asgMock, iamMock, eksMock, ec2Mock)
 	ctx := MockContext(ig, k, w)
 	state := ctx.GetDiscoveredState()
-	state.SetCluster(&eks.Cluster{
-		Version: aws.String("1.15"),
-		CertificateAuthority: &eks.Certificate{
-			Data: aws.String("clusterca"),
-		},
-		Endpoint: aws.String("foo.amazonaws.com"),
-	})
+	state.SetCluster(MockEksCluster("1.15"))
 	state.Publisher.Client = k.Kubernetes
 	state.ScalingConfiguration = &scaling.LaunchConfiguration{
 		AwsWorker: w,
@@ -142,12 +135,7 @@ func TestCreateScalingGroupPositive(t *testing.T) {
 		Publisher: kubeprovider.EventPublisher{
 			Client: k.Kubernetes,
 		},
-		Cluster: &eks.Cluster{
-			CertificateAuthority: &eks.Certificate{
-				Data: aws.String("foo"),
-			},
-			Endpoint: aws.String("foo.amazonaws.com"),
-		},
+		Cluster:              MockEksCluster(""),
 		ScalingConfiguration: lc,
 	})
 
@@ -187,13 +175,7 @@ func TestCreateNoOp(t *testing.T) {
 	}
 
 	ctx.SetDiscoveredState(&DiscoveredState{
-		Cluster: &eks.Cluster{
-			CertificateAuthority: &eks.Certificate{
-				Data: aws.String(""),
-			},
-			Endpoint:           aws.String("foo.amazonaws.com"),
-			ResourcesVpcConfig: &eks.VpcConfigResponse{},
-		},
+		Cluster: MockEksCluster(""),
 		Publisher: kubeprovider.EventPublisher{
 			Client: k.Kubernetes,
 		},
--- a/controllers/provisioners/eks/eks.go
+++ b/controllers/provisioners/eks/eks.go
@@ -17,6 +17,7 @@ package eks
 
 import (
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
 	"sync"
 
 	"github.com/go-logr/logr"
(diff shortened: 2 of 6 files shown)


controllers/provisioners/eks/update_test.go
```diff
@@ -109,7 +109,12 @@ func TestUpdateWithDriftRotationPositive(t *testing.T) {
 		},
 		ClusterNodes: nodes,
 		Cluster: &eks.Cluster{
-			Version: aws.String("1.15"),
+			CertificateAuthority: &eks.Certificate{
```
eytan-avisror, line 112: Can we create a MockEksCluster function with default values and then have a more tidy declaration? e.g.
```golang
Cluster: MockEksCluster()
```
you can add any element we want to control as arguments to this function.
This will remove all the dupe code

controllers/provisioners/eks/helpers.go
```diff
@@ -350,34 +380,49 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
 		isOverride = true
 		overrideLabels := strings.Split(val, ",")
 		for _, label := range overrideLabels {
-			labelList = append(labelList, label)
+			keyVal := strings.Split(label, "=")
```
eytan-avisror, line 383: what happens if label provided has no value? should have safety here around accessing `keyVal`?

controllers/provisioners/eks/helpers.go
```diff
@@ -350,34 +380,49 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
...
-		labelList = append(labelList, fmt.Sprintf(RoleNewLabelFmt, instanceGroup.GetName()))
+		labelMap[RoleNewLabel] = instanceGroup.GetName()
 
 		// add the old style role label if the cluster's k8s version is < 1.16
 		clusterVersion := ctx.DiscoveredState.GetClusterVersion()
 		ver, err := semver.NewVersion(clusterVersion)
 		if err != nil {
 			ctx.Log.Error(err, "Failed parsing the cluster's kubernetes version", "instancegroup", instanceGroup.GetName())
-			labelList = append(labelList, fmt.Sprintf(RoleOldLabelFmt, instanceGroup.GetName()))
+			labelMap[fmt.Sprintf(RoleOldLabel, instanceGroup.GetName())] = ""
```
eytan-avisror, line 397: Looks like we are changing the Role old label here from:
`"node-role.kubernetes.io/role=\"\""` to `"node-role.kubernetes.io/role"`
Should we use RoleOldLabelFmt?

controllers/provisioners/eks/helpers.go
```diff
@@ -350,34 +380,49 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
...
 		} else {
 			c, _ := semver.NewConstraint("< 1.16-0")
 			if c.Check(ver) {
-				labelList = append(labelList, fmt.Sprintf(RoleOldLabelFmt, instanceGroup.GetName()))
+				labelMap[fmt.Sprintf(RoleOldLabel, instanceGroup.GetName())] = ""
```
eytan-avisror, line 401: Same here, RoleOldLabelFmt to RoleOldLabel

controllers/provisioners/eks/helpers.go
eytan-avisror: Nevermind - I guess since this is a map we no longer need the value

controllers/provisioners/eks/helpers.go
```diff
@@ -350,34 +380,49 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
...
+func (ctx *EksInstanceGroupContext) GetLabelList() []string {
+	var (
+		labelList []string
+	)
+
+	for key, value := range ctx.GetComputedLabels() {
+		if value == "" {
+			value = "\"\""
+		}
+		labelList = append(labelList, fmt.Sprintf("%v=%v", key, value))
```
eytan-avisror, line 424: What happens if value is `""`? is it rejected by API?

controllers/provisioners/eks/helpers.go
```diff
@@ -97,18 +97,43 @@ func (ctx *EksInstanceGroupContext) ResolveSecurityGroups() []string {
...
 <powershell>
   [string]$EKSBinDir = "$env:ProgramFiles\Amazon\EKS"
   [string]$EKSBootstrapScriptName = 'Start-EKSBootstrap.ps1'
   [string]$EKSBootstrapScriptFile = "$EKSBinDir\$EKSBootstrapScriptName"
   & $EKSBootstrapScriptFile -EKSClusterName {{ .ClusterName }} -KubeletExtraArgs '{{ .KubeletExtraArgs }}' 3>&1 4>&1 5>&1 6>&1
-</powershell>
-		`
-	} else {
+</powershell>`
+	case OsFamilyBottleRocket:
```
eytan-avisror, line 121: Looks like we added apiServer & certificate into our context, where in the case of linux the bootstrap.sh script polls for this.
Should we align all OS families to same approach?
Whether we query in userdata or if we use the state we collected does not matter as long as it does not cost in additional API calls - so as long as we already have the data (via DescribeCluster), should we pass the api / certificate into bootstrap.sh script? or the opposite - should we make the bottlerocket/other userdata query those values?
https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L223-L241

controllers/provisioners/eks/eks_test.go
```diff
@@ -92,13 +96,39 @@ func MockKubernetesClientSet() kubeprovider.KubernetesClientSet {
...
 		InstanceGroup: instanceGroup,
 		Log:           ctrl.Log.WithName("unit-test").WithName("InstanceGroup"),
 	}
-	return New(input)
+	context := New(input)
+
+	context.DiscoveredState = &DiscoveredState{
+		Provisioned:          false,
+		NodesReady:           false,
+		ClusterNodes:         nil,
```
eytan-avisror, line 114: we can probably remove all the fields which are nil as that would be the default.

controllers/provisioners/eks/eks.go
```diff
@@ -34,13 +35,17 @@ const (
 	OsFamilyAnnotation                  = "instancemgr.keikoproj.io/os-family"
 	ClusterAutoscalerEnabledAnnotation  = "instancemgr.keikoproj.io/cluster-autoscaler-enabled"
 
-	OsFamilyWindows                     = "windows"
+	OsFamilyWindows      = "windows"
+	OsFamilyBottleRocket = "bottlerocket"
```
eytan-avisror, line 39: Should we add amazonlinux2 annotation and make it the default value?
e.g. in validation of the API (instancegroups_types.go) can add this annotation if it's missing? or should we just assume that if this annotation is missing, it's linux - WDYT?

controllers/provisioners/eks/create_test.go
```diff
@@ -46,7 +46,13 @@ func TestCreateManagedRolePositive(t *testing.T) {
 	w := MockAwsWorker(asgMock, iamMock, eksMock, ec2Mock)
 	ctx := MockContext(ig, k, w)
 	state := ctx.GetDiscoveredState()
-	state.SetCluster(&eks.Cluster{Version: aws.String("1.15")})
+	state.SetCluster(&eks.Cluster{
```
eytan-avisror, line 49: You can use `MockEksCluster()` here as well if you add a function for it

controllers/provisioners/eks/helpers.go
backjo: The reason we need this is that BottleRocket doesn't support querying these values - they're required input parameters for UserData. I think it makes sense to continue allowing the bootstrap script in the AL2 and Windows AMIs to look it up.

controllers/provisioners/eks/eks.go
backjo: Yeah, that probably makes sense.

controllers/provisioners/eks/helpers.go
backjo: This conditional was to provide compatibility with existing behavior under test. I'm not sure w.r.t. it being rejected.

controllers/provisioners/eks/helpers.go
backjo: Though, one argument against that is that the AMIs aren't always consistent with each other - see (https://github.com/bottlerocket-os/bottlerocket/issues/1197). Injecting information from instance-manager could help protect against AMI drift/inconsistency.

controllers/provisioners/eks/eks.go
backjo: Added this.

controllers/provisioners/eks/helpers.go
eytan-avisror: We can probably keep this as is, might want to pass the same flags to bootstrap.sh (since it accepts these in flags as override) in the future - it will make bootstrapping faster and more reliable.

controllers/provisioners/eks/helpers.go
backjo: Added len check for this slice to ensure we set it to "" if there are not 2 values in the slice.

controllers/provisioners/eks/eks.go
```diff
@@ -114,13 +124,13 @@ func (ctx *EksInstanceGroupContext) GetInstanceGroup() *v1alpha1.InstanceGroup {
...
-		instanceGroup  = ctx.GetInstanceGroup()
-		annotations = instanceGroup.GetAnnotations()
+		instanceGroup = ctx.GetInstanceGroup()
+		annotations   = instanceGroup.GetAnnotations()
 	)
 	if _, exists := annotations[OsFamilyAnnotation]; exists {
 		return annotations[OsFamilyAnnotation]
 	}
-	return "default"  // return "" is will also be fine here
+	return OsFamilyAmazonLinux2 // return "" is will also be fine here
```
eytan-avisror, line 133: I think we can also remove the comment 😅  - I think it was a review comment that went in with the suggestion

backjo: lol - yep

controllers/provisioners/eks/helpers.go
eytan-avisror: mm, I don't see it - did you push it?

eytan-avisror: Also, what if someone creates a malformed annotation (e.g. no equal sign)?

eytan-avisror: Since this is user input we should be careful with out of bounds index

backjo: strings.split will return a slice with the original string if the seperator does not exist.

eytan-avisror: @backjo can you point me to the len check? 😄

eytan-avisror: nvm, see it now 😓

eytan-avisror (approved):
Awesome 🥇 

Can you just do one quick test?
When you have an existing instance group (created with old controller), upgrade to latest controller and make sure there is no node rotation.

This is just to avoid a release the causes a node rotation - sometimes when we change code around labels / userData this can happen when something minor changes.

backjo:
@eytan-avisror testing with an IG from master, running instance-manager on the latest commit here did not change the status of the IG, which remained in Ready

Issue #198, opened 2020-11-15
Add support for AWS Launch Template placement configuration


preflightsiren:
After supporting #17 Launch Templates can be used to support many more valid configurations.
One configuration that is super-important to me is Dedicated Hosts support which would allow ec2 instances to be automatically placed onto dedicated hardware.

For minimal support we need:

from the AWS SDK `ec2/api.go`:

```
// The information to include in the launch template.
type RequestLaunchTemplateData struct {
	... snip ...

	// The license configurations.
	LicenseSpecifications []*LaunchTemplateLicenseConfigurationRequest `locationName:"LicenseSpecification" locationNameList:"item" type:"list"`

	... snip ...

	// The placement for the instance.
	Placement *LaunchTemplatePlacementRequest `type:"structure"`

	... snip ...
}
```

```
type LaunchTemplateLicenseConfigurationRequest struct {
	_ struct{} `type:"structure"`

	// The Amazon Resource Name (ARN) of the license configuration.
	LicenseConfigurationArn *string `type:"string"`
}
```

After some trial and error, these are the only fields configurable for AutoScalingGroups.
```
// Describes the placement of an instance.
type LaunchTemplatePlacementRequest struct {

	// The Availability Zone for the instance.
	AvailabilityZone *string `type:"string"`

	// The ARN of the host resource group in which to launch the instances. If you
	// specify a host resource group ARN, omit the Tenancy parameter or set it to
	// host.
	HostResourceGroupArn *string `type:"string"`

	// The tenancy of the instance (if the instance is running in a VPC). An instance
	// with a tenancy of dedicated runs on single-tenant hardware.
	Tenancy *string `type:"string" enum:"Tenancy"`
}
```

eytan-avisror:
Completed with #199

Issue #193, opened 2020-11-11
Documentation: Better README.md


eytan-avisror:
We should make our main readme more concise and possible include a screen capture GIF to show a quick demo of instance-manager.

We should remove the more in-depth technical details into their own documents possibly in docs folder.

PR #199, opened 2020-11-15
Launch template placement support

preflightsiren:
Implementation for #198 
Also contributes to #193


controllers/provisioners/eks/helpers.go
```diff
@@ -890,7 +891,7 @@ func (ctx *EksInstanceGroupContext) GetOverrides() []*autoscaling.LaunchTemplate
 	)
 	overrides := []*autoscaling.LaunchTemplateOverrides{}
 
-	if mixedPolicy.InstanceTypes != nil {
+	if mixedPolicy != nil && mixedPolicy.InstanceTypes != nil {
```
preflightsiren, line 894: Mentioned this in slack; I hit a case pretty early on where `mixedPolicy` was `nil` - maybe resolved with the latest changes.

preflightsiren:
Screen Shot 2020-11-02 at 7 34 36 pm
Tested locally.

Recommendations welcome where we can improve the testing structure

preflightsiren:
TODO:
- [ ] Update examples
- [ ] Update changelog

controllers/provisioners/eks/helpers.go
eytan-avisror: mixedPolicy is guaranteed to be non nil inside GetOverrides due to:
https://github.com/keikoproj/instance-manager/blob/master/controllers/provisioners/eks/helpers.go#L930-L934

controllers/provisioners/eks/scaling/launchtemplate.go
```diff
@@ -347,3 +406,12 @@ func sortVersions(versions []*ec2.LaunchTemplateVersion) []*ec2.LaunchTemplateVe
 
 	return versions
 }
+
+func contains(src []string, value string) bool {
```
eytan-avisror, line 410: You can use `common.ContainsEqualFold` and remove this

controllers/provisioners/eks/scaling/launchtemplate.go
```diff
@@ -311,6 +349,27 @@ func (lt *LaunchTemplate) blockDeviceList(volumes []v1alpha1.NodeVolume) []*ec2.
 	return sortTemplateDevices(devices)
 }
 
+func launchTemplateLicenseSpeficicationRequest(s []string) []*ec2.LaunchTemplateLicenseConfigurationRequest {
```
eytan-avisror, line 352: Can we move both `launchTemplateLicenseSpeficicationRequest` and `launchTemplatePlacementRequest` into awsprovider? 
it might be more appropriate there

api/v1alpha1/instancegroup_types.go
```diff
@@ -187,26 +187,28 @@ type EKSSpec struct {
...
+	Labels                      map[string]string            `json:"labels,omitempty"`
+	Taints                      []corev1.Taint               `json:"taints,omitempty"`
+	UserData                    []UserDataStage              `json:"userData,omitempty"`
+	ExistingRoleName            string                       `json:"roleName,omitempty"`
+	ExistingInstanceProfileName string                       `json:"instanceProfileName,omitempty"`
+	ManagedPolicies             []string                     `json:"managedPolicies,omitempty"`
+	MetricsCollection           []string                     `json:"metricsCollection,omitempty"`
+	LifecycleHooks              []LifecycleHookSpec          `json:"lifecycleHooks,omitempty"`
+	MixedInstancesPolicy        *MixedInstancesPolicySpec    `json:"mixedInstancesPolicy,omitempty"`
+	LicenseSpecifications       []string                     `json:"licenseSpecifications,omitempty"`
```
eytan-avisror, line 210: Are both theses features `LicenseSpecifications` and `Placement` support by both LaunchTemplate and LaunchConfig?

api/v1alpha1/instancegroup_types.go
```diff
@@ -487,6 +495,12 @@ func (c *EKSConfiguration) Validate() error {
 			return err
 		}
 	}
+
+	for i, v := range c.LicenseSpecifications {
+		if !common.StringEmpty(v) && !strings.HasPrefix(v, awsprovider.ARNPrefix) {
```
eytan-avisror, line 500: Should this be || instead?

api/v1alpha1/instancegroup_types.go
```diff
@@ -187,26 +187,28 @@ type EKSSpec struct {
...
+	Taints                      []corev1.Taint               `json:"taints,omitempty"`
+	UserData                    []UserDataStage              `json:"userData,omitempty"`
+	ExistingRoleName            string                       `json:"roleName,omitempty"`
+	ExistingInstanceProfileName string                       `json:"instanceProfileName,omitempty"`
+	ManagedPolicies             []string                     `json:"managedPolicies,omitempty"`
+	MetricsCollection           []string                     `json:"metricsCollection,omitempty"`
+	LifecycleHooks              []LifecycleHookSpec          `json:"lifecycleHooks,omitempty"`
+	MixedInstancesPolicy        *MixedInstancesPolicySpec    `json:"mixedInstancesPolicy,omitempty"`
+	LicenseSpecifications       []string                     `json:"licenseSpecifications,omitempty"`
+	Placement                   *LaunchTemplatePlacementSpec `json:"placement,omitempty"`
```
eytan-avisror, line 211: If `Placement` can be supported by LaunchConfiguration as well, we should use a more generic name e.g. `ScalingConfigurationPlacementSpec`

controllers/provisioners/eks/scaling/launchtemplate.go
```diff
@@ -243,6 +246,41 @@ func (lt *LaunchTemplate) Drifted(input *CreateConfigurationInput) bool {
 		drift = true
 	}
 
+	if len(latestVersion.LaunchTemplateData.LicenseSpecifications) != len(input.LicenseSpecifications) {
```
eytan-avisror, line 249: It might be eaiser to create a `ec2.LaunchTemplateLicenseConfiguration` object from `input.LicenseSpecifications`, make sure both are sorted, and then just do a reflect.DeepEqual() - same as how we do Volumes - this will also allow us to remove the len check.

controllers/provisioners/eks/scaling/launchtemplate.go
```diff
@@ -243,6 +246,41 @@ func (lt *LaunchTemplate) Drifted(input *CreateConfigurationInput) bool {
...
+		if !contains(input.LicenseSpecifications, aws.StringValue(v.LicenseConfigurationArn)) {
+			log.Info("detected drift", "reason", "LicenseSpecifications has changed", "instancegroup", lt.OwnerName,
+				"previousValue", aws.StringValue(v.LicenseConfigurationArn),
+				"newValue", input.LicenseSpecifications,
+			)
+			drift = true
+		}
+	}
+
+	if input.Placement == nil {
```
eytan-avisror, line 263: Same, we should have a function to get a `LaunchTemplatePlacement` from `input.Placement`  in the AWS pkg which does this nil check as well.
Then all we need is sorting and reflect.DeepEqual()

api/v1alpha1/instancegroup_types.go
```diff
@@ -224,6 +226,12 @@ type MixedInstancesPolicySpec struct {
 	InstanceTypes []*InstanceTypeSpec `json:"instanceTypes,omitempty"`
 }
 
+type LaunchTemplatePlacementSpec struct {
+	AvailabilityZone     string `json:"availabilityZone,omitempty"`
+	HostResourceGroupArn string `json:"hostResourceGroupArn,omitempty"`
+	Tenancy              string `json:"tenancy,omitempty"`
```
eytan-avisror, line 232: Are these the minimal values required to define placement? Are they all required? is it possible to just provide `Tenancy` or other fields?

api/v1alpha1/instancegroup_types.go
preflightsiren: its based upon the logic above: https://github.com/keikoproj/instance-manager/pull/199/files/03b5e05aa1abae87f64cc14bcc05154451ab9cf9#diff-9f09697a5f636c8e5c9996db1dac259666efd6f1631577aa6f4bb814acd685f8R447

we could flip the logical statement to:

`if common.StringEmpty(v) || strings.HasPrefix(v, awsprovider.ARNPrefix) { //pass }` 

The way I would express it in English is; the LicenseSpecifications array must only contain empty strings, or valid ARNs.

controllers/provisioners/eks/helpers.go
preflightsiren: not sure I'm following. `mixedPolicy` comes from `configuration.GetMixedInstancesPolicy()`

https://github.com/keikoproj/instance-manager/blob/6f9289fecaf6793a42a399cf902e1f18670082d0/api/v1alpha1/instancegroup_types.go#L593-L595

which can be nil?

api/v1alpha1/instancegroup_types.go
preflightsiren: Unfortunately `Placement` isn't supported by LaunchConfiguration.

api/v1alpha1/instancegroup_types.go
preflightsiren: Unfortunately no, they're specific to LaunchTemplates.

api/v1alpha1/instancegroup_types.go
preflightsiren: These are the only 3 of 8 fields _supported_ by AutoScalingGroups.
`AvailabilityZone` is optional.
`Tenancy` must be `host` if `HostResourceGroupArn` is set; otherwise they're completely optional.

controllers/provisioners/eks/helpers.go
eytan-avisror: Correct, but `GetDesiredMixedInstancesPolicy` calls `GetOverrides` and does a nil check.
Eitherway, this can probably stay / won't hurt.

controllers/provisioners/eks/scaling/launchtemplate.go
preflightsiren: doing seems to create a circular dependency:

```
package command-line-arguments
        imports github.com/keikoproj/instance-manager/api/v1alpha1
        imports github.com/keikoproj/instance-manager/controllers/providers/aws
        imports github.com/keikoproj/instance-manager/controllers/provisioners/eks/scaling
        imports github.com/keikoproj/instance-manager/api/v1alpha1: import cycle not allowed
package command-line-arguments
        imports github.com/keikoproj/instance-manager/api/v1alpha1
        imports github.com/keikoproj/instance-manager/controllers/providers/aws
        imports github.com/keikoproj/instance-manager/controllers/provisioners/eks/scaling
        imports github.com/keikoproj/instance-manager/controllers/providers/aws: import cycle not allowed
```

api/v1alpha1/instancegroup_types.go
eytan-avisror: Looks like it is supported according to:
https://docs.aws.amazon.com/sdk-for-go/api/service/autoscaling/#CreateAutoScalingGroupInput

Should we have a more generic implementation in this case?

controllers/provisioners/eks/scaling/launchtemplate.go
eytan-avisror: This is because of `LaunchTemplatePlacementInput` which is part of scaling package, suggest you remove that struct and just pass in the input you need as arguments.

e.g.
```go
func (w *AwsWorker) GetLaunchTemplatePlacementRequest(az, arn, tenancy string) {
  ...
}
```

api/v1alpha1/instancegroup_types.go
eytan-avisror: What does it mean if it contains an empty string?
Isn't an empty string considered invalid? it would mean someone provided it as:
```yaml
configuration:
  licenseSpecifications:
  - ""
```
In that case, shouldnt it be: `the LicenseSpecifications array must only contain valid ARNs.`

api/v1alpha1/instancegroup_types.go
eytan-avisror: In that case, we should probably add validations accordingly for both `Tenancy` (can only be one of allowed values), and `HostResourceGroupArn` (tenancy must be `host` and value must be ARN).

controllers/provisioners/eks/scaling/launchtemplate.go
```diff
@@ -311,6 +349,27 @@ func (lt *LaunchTemplate) blockDeviceList(volumes []v1alpha1.NodeVolume) []*ec2.
...
+	}
+	return output
+}
+
+func launchTemplatePlacementRequest(input *LaunchTemplatePlacementInput) *ec2.LaunchTemplatePlacementRequest {
+	if input == nil {
+		return &ec2.LaunchTemplatePlacementRequest{}
+	}
+	return &ec2.LaunchTemplatePlacementRequest{
+		AvailabilityZone:     aws.String(input.AvailabilityZone),
```
eytan-avisror, line 367: Since this is an optional field - it might be problematic to set it.
We should only set `AvailabilityZone` if it is not empty.

api/v1alpha1/instancegroup_types.go
preflightsiren: yeah that's fair.

api/v1alpha1/instancegroup_types.go
preflightsiren: ahh, `PlacementTenancy` is a subset of `LaunchTemplatePlacement` it's essentially the equivalent of `LaunchTemplatePlacement.tenancy`.

Are there use cases to support placement tenancy in LaunchConfigurations?

eytan-avisror: Not currently, but in case there will be in the future - just wondering if we should have a more generic naming for the API, e.g. instead of `LaunchTemplatePlacementSpec`, just use `PlacementSpec`.
We can possibly make `.tenancy` support for launch configurations now, or down the road

preflightsiren: given the `PlacementSpec` has extra configurable fields that are only appropriate for LaunchTemplates, what would support for LT's and LC's both look like?

My read of your suggestion is that we would rename `LaunchTemplatePlacementSpec` -> `PlacementSpec`, and add validation that when `type == LaunchConfiguration` fail validation for any extra fields. is that what you're thinking?

preflightsiren: I've wired up the `PlacementTenancy` for `LaunchConfiguration`s but having some trouble validating the `EKSConfiguration` to make provide meaningful feedback depending on LC vs LT - do you have a suggestion on how we should validate this?

preflightsiren:
I've addressed all the feedback and have provided updated docs and tests. Ready for review and final feedback.

README.md
```diff
@@ -361,6 +368,29 @@ spec:
 
 if you do not provide these fields, a role will be created for your instance-group by the controller (will require IAM access).
 
+
+#### EC2 Instance placement
```
eytan-avisror, line 372: We should probably avoid adding more examples to main readme.md, we also have a ticket to make it shorter #193 
Wonder if we should add a file docs/examples/EKS.md and place all the examples there?

controllers/provisioners/eks/scaling/interface.go
```diff
@@ -58,6 +58,14 @@ type CreateConfigurationInput struct {
 	Volumes               []v1alpha1.NodeVolume
 	UserData              string
 	SpotPrice             string
+	LicenseSpecifications []string
+	Placement             *PlacementInput
```
eytan-avisror, line 62: I wondering if we should just use `v1alpha1.Placement` directly instead of having another struct `PlacementInput` here?
Sane as we do for Volumes in line 58. This would also make `GetPlacementInput` redundant
WDYT?

docs/EKS.md
```diff
@@ -215,6 +219,22 @@ spec:
         effect: <string> : the effect of the taint
 ```
 
+### PlacementSpec
+
+Represents the EC2 Placement information for your EC2 instances.
+All fields are supported for Launch Templates; Launch Configuration's only support setting `tenancy`.
```
eytan-avisror, line 225: Should we also add another example such as the one below to show the usage in case of LaunchConfiguration.

docs/EKS.md
```diff
@@ -215,6 +219,22 @@ spec:
...
+
+```yaml
+spec:
+  provisioner: eks
+  eks:
```
eytan-avisror, line 230: Should we also add `type: LaunchTemplate` so it's not ambiguous ?

eytan-avisror:
Great stuff @preflightsiren , very close to merging it, just had few more questions / comments

controllers/provisioners/eks/scaling/interface.go
preflightsiren: Sounds good, I forget where but I remember we used an internal model in one other case. The only argument I can see is that we might want to change the internal API independent of the public API.

I'll try it out with `v1alpha.PlacementSpec` and see what it looks like :)

eytan-avisror (approved):
Awesome!! Will run functional test and merge this as soon as it passes

Issue #137, opened 2020-06-07
Move from Travis CI to Github Actions


eytan-avisror:
Github Actions seem to be much better and faster than Travis, and allows up to 6hr jobs.
Should explore moving our CI to Github Actions.

Need:
- [x]  Run unit-test & codecov on every pull request commit
- [x] When tag is pushed, publish an image (release)
- [x] When label / comment is added 'ok-to-test' run functional test (block concurrency)
- [x] cron based functional test (nightly)

PR #213, opened 2020-11-24
Add Github Actions for unit test and image publication

### LINKED ISSUE #191: Support ARM architecture
Feature Request
**What happened**:
ARM-based processors are becoming increasingly popular. We should leverage the cross-platform capabilities in Docker to build and publish images for ARM64 in addition to AMD64

backjo:
Closes #191 
Partially implements #137

This PR adds two GitHub actions - one for running unit tests, and another for publishing images. 

The image publication adds several new features above and beyond the existing image publication today:
- MultiArch images are now published (see [example](https://hub.docker.com/layers/jbacklegalzoom/instancemgr/pr-1/images/sha256-502fb4cd61290eb91c2cc9f63794c971b2cf73859eee2770d2222359411d66c4?context=repo))
- PR images are automatically published
- Images can easily be published to multiple repos - this PR adds GitHub container registry, which does not have any limits on downloading containers. 
 

Before any merge, this would needs secrets added to this repo for pushing to GHCR and DockerHub

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,83 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+    tags:
+    - v*
+  pull_request:
+    branches: [ master ]
+  
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
+  
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Test
+      run: |
+        go test -v ./controllers/... -coverprofile coverage.txt
+    - name: Upload to Codecov
+      uses: codecov/codecov-action@v1
+      with:
+        file: ./coverage.txt
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+      with:
+        install: true
+        version: latest
+        
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+      with:
+        image: tonistiigi/binfmt:latest
+        platforms: all
+  
+    - name: Login to DockerHub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+ 
+    -
+      name: Login to GitHub Container Registry
+      uses: docker/login-action@v1 
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.CR_PAT }}
+
+    -
+      name: Docker meta
+      id: docker_meta
+      uses: crazy-max/ghaction-docker-meta@v1
+      with:
+        images: ${{ github.repository_owner }}/instancemgr,ghcr.io/${{ github.repository_owner }}/instance-manager
+    -
+      name: Build and push
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./Dockerfile
+        platforms: linux/amd64,linux/arm/v7,linux/arm64
+        push: true
+        tags: ${{ steps.docker_meta.outputs.tags }}
+       
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ master ]
     tags:
-    - v*
+    - *
   pull_request:
     branches: [ master ]
   
@@ -42,7 +42,7 @@ jobs:
       uses: docker/login-action@v1 
       with:
         registry: ghcr.io
-        username: ${{ github.repository_owner }}
+        username: ${{ secrets.CR_USERNAME }}
         password: ${{ secrets.CR_PAT }}
 
     -
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -9,26 +9,6 @@ on:
     branches: [ master ]
   
 jobs:
-  test:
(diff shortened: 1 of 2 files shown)


eytan-avisror:
Thanks @backjo 
Looking forward to the day we can get rid of Travis 🙏  🙏  it has become incredibly slow to process builds.

OOO today, but will review this tomorrow  !

.github/workflows/push.yml
```diff
@@ -0,0 +1,63 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+    tags:
+    - v*
```
eytan-avisror, line 7: I think the tags we usually push on release don't have a `v` in front of them, so if this is meant to trigger on that it won't work.

.github/workflows/push.yml
```diff
@@ -0,0 +1,63 @@
...
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+ 
+    -
+      name: Login to GitHub Container Registry
+      uses: docker/login-action@v1 
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.CR_PAT }}
```
eytan-avisror, line 46: I assume we need to set up an account before this will work? or does this tie in to the repo identity already?

.github/workflows/push.yml
```diff
@@ -0,0 +1,63 @@
...
+        images: ${{ github.repository_owner }}/instancemgr,ghcr.io/${{ github.repository_owner }}/instance-manager
+    -
+      name: Build and push
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./Dockerfile
+        platforms: linux/amd64,linux/arm/v7,linux/arm64
+        push: true
+        tags: ${{ steps.docker_meta.outputs.tags }}
```
eytan-avisror, line 62: So which tags will this actually push?
Our current CI pushes an image with `master` tag on every merge to master branch, and on every release tag, it pushes the tag name as the image tag e.g. `0.9.0`

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+  
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.15
```
eytan-avisror, line 17: I think we are still using go 1.13 in go.mod
https://github.com/keikoproj/instance-manager/blob/master/go.mod#L3

Not sure if we want to keep these the same for now

eytan-avisror:
@backjo 
Looks great, just had a few questions / comments.
Also, I think the only gap to get rid of travis is the BDD test - if we can move that one too we should be able to get rid of the travisfile we have and completely move to github actions, but not sure how trivial it is to do, so potentially we can do that later.
The BDD test runs nightly on a cron schedule, deploys latest master branch, and runs `make bdd`..
see here https://github.com/keikoproj/instance-manager/blob/master/.travis.yml#L49-L60

we can probably disable travis once we merge this but keep the travis file until we can verify all test / push actions are working as expected

.github/workflows/push.yml
backjo: The account should already be setup, but we'll need someone with access to this repo to generate a Personal Access Token with write:packages permissions to be able to publish to it.

.github/workflows/push.yml
backjo: From the GH Action docs:

Event | Ref | Commit SHA | Docker Tags
-- | -- | -- | --
pull_request | refs/pull/2/merge | a123b57 | pr-2
push | refs/heads/master | cf20257 | master
push | refs/heads/my/branch | a5df687 | my-branch
push tag | refs/tags/v1.2.3 | ad132f5 | v1.2.3, latest
push tag | refs/tags/v2.0.8-beta.67 | fc89efd | v2.0.8-beta.67, latest

backjo:
Yeah, I was thinking of doing the BDD later - haven't even run that locally yet myself, so figured it'd take a bit more time

.github/workflows/push.yml
eytan-avisror: Cool, I can probably do that before merging

.github/workflows/push.yml
eytan-avisror: ok nice, so if I understand correctly, on master merge it would push `master` and on tag push it would push same tag name + `latest`.

eytan-avisror:
> Yeah, I was thinking of doing the BDD later - haven't even run that locally yet myself, so figured it'd take a bit more time

Sounds good, should we comment these steps out on travis in this case?
Or should we have both run for now? WDYT
Should not really matter, but one pipeline would re-push same image tags if we don't comment out

.github/workflows/push.yml
```diff
@@ -0,0 +1,63 @@
...
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+ 
+    -
+      name: Login to GitHub Container Registry
+      uses: docker/login-action@v1 
+      with:
+        registry: ghcr.io
+        username: ${{ secrets.CR_USERNAME }}
```
eytan-avisror, line 45: Should this actually be `${{ github.repository_owner }}` according to https://github.com/docker/login-action#github-container-registry 

Also, should I create the PAT and add the token as a secret named CR_PAT?
I already added organization level secrets for DOCKERHUB_USERNAME & DOCKERHUB_TOKEN

eytan-avisror: nvm, I understand, it should be the user that created the secret, in this case my username.
I will add a secret for CR_USERNAME as well

eytan-avisror (approved):
💯

.github/workflows/push.yml
backjo: yep

.github/workflows/push.yml
backjo: yep, that's why i switched it

eytan-avisror:
@backjo one of the workflows failed due to the tags being `*` - I guess that is not supported?

PR #214, opened 2020-11-30
Fix premature upgrade completion

### LINKED ISSUE #211: crd strategy - reconcile may complete before resource is completed
Looks like there is a code path where we can mark reconcile successful prematurely during an upgrade.

According to the following snippet:
https://github.com/keikoproj/instance-manager/blob/d84908b7bb5a0b4d341b52e27b4f5aff43e21c0e/controllers/provisioners/eks/update.go#L107-L115

It seems we don't enter the upgrade path nodes are ready and rotationNeeded is not true - In this case it can happen when the last node is rotated (as now rotationNeeded is false), however, the upgrade resource object is still running.

rotationNeeded should also check if the `status.strategyResourceName` resource is still active.

eytan-avisror:
Signed-off-by: Eytan Avisror <Eytan_Avisror@hotmail.com>

Fixes #211 

This marks rotationNeeded if the custom resource for node rotation is still active.
Without this, we are exposed to a race condition where if during an upgrade all nodes are ready we may mark it completed prematurely.

This requires adding another status field to cache the upgrade resource's namespace

TBD:
- [x] functional test

### DIFF
--- a/api/v1alpha1/instancegroup_types.go
+++ b/api/v1alpha1/instancegroup_types.go
@@ -309,6 +309,7 @@ type InstanceGroupStatus struct {
 	ActiveScalingGroupName        string                   `json:"activeScalingGroupName,omitempty"`
 	NodesArn                      string                   `json:"nodesInstanceRoleArn,omitempty"`
 	StrategyResourceName          string                   `json:"strategyResourceName,omitempty"`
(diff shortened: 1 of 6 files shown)


tekenstam (approved):

kevdowney (approved):
LGTM

gaganapplatix (approved):
LGTM

PR #215, opened 2020-12-01
Fix special character in GH Workflow

backjo:

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -5,8 +5,6 @@ on:
     branches: [ master ]
     tags:
     - '*'
-  pull_request:
-    branches: [ master ]
   
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


eytan-avisror (approved):

eytan-avisror:
@backjo Any idea why push is failing?
WF says:
```
Run docker/login-action@v1
Error: Username and password required
```
following secrets exist:
Screen Shot 2020-12-01 at 3 48 30 PM

backjo:
I think it's a security feature - at least for this PR. It sounds like GH actions doesn't allow access to secrets from forks. Could you push this branch to this repo and create a PR off of it?

eytan-avisror:
> It sounds like GH actions doesn't allow access to secrets on PRs from forks.

I think you're right - I've had the same issue in the past while trying out github actions and came to same conclusion. 
In this case, should we just remove the PR trigger and stick with push only?

backjo:
Yep, let's do it.

backjo:
@eytan-avisror when you get a chance, can you make the GHCR repository public? It defaults to private.

eytan-avisror:
@backjo fixed

Issue #184, opened 2020-11-02
EKS provisioner should support --use-max-pods field in spec


backjo:
**Is this a BUG REPORT or FEATURE REQUEST?**:
Feature Request
**What happened**:
If using custom networking on EKS, the number of max pods per node is reduced (see https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html). 

Instance Manager should allow setting --use-max-pods to false, so that nodes using custom networking can correctly set the maximum number of pods in bootstrap args. (Today, bootstrap args goes directly into --kubelet-extra-args)

eytan-avisror:
Should we support this through annotations? e.g. `instancemgr.keikoproj.io/kubelet-arguments/use-max-pods: "false"`

I think changing the `bootstrapArgs` field to take kubelet arguments vs. --kubelet-extra-args is a bigger deal since it will break backwards compatiblity.

There are potentially other kubelet arguments we would like to allow control over.
WDYT

backjo:
@eytan-avisror totally think we should not break backwards compatibility, and 'bootstrapArguments' should still continue to be what we pass into --kubelet-extra-args

I think one thing we should take into account is how these settings might map to BottleRocket AMIs - given that a lot of these properties overlap, but they are injected via TOML-formatted data in user data. 

I wonder if it makes sense to maintain bootstrapArguments as an escape hatch for new/unsupported arguments, but add some sort of structure for EKS-specific settings that might need special mapping or handling, similar to how node labels and taints are specified as part of the spec today. 
i.e. something like 
```yaml
configuration:
  bootstrapArgs: "--foo=bar"
  bootstrapData:
    maxPods: 20
    kubeletExtraArgs:
      evictionHard: "memory.available<300Mi"
```

This would map to the following bootstrap script for AL2 AMIs
```bash
bootstrap.sh --use-max-pods 'false' --kubelet-extra-args '--max-pods "20" --eviction-hard "memory.available<300mi" --foo="bar"'
```
But end up looking like this for BottleRocket AMIs
```toml
[settings.kubernetes]
max-pods = 20
```

Given that fields like max-pods are part of the core EKS bootstrap script or properties file, I think it makes more sense that they live as part of spec rather than as an annotation, but that's not a strong opinion.

eytan-avisror:
So there are two types of flags here:
- Ones that go directly to the bootstrap.sh script such as `--use-max-pods`
- Ones that go to kubelet via `--kubelet-extra-args`

[...]

PR #216, opened 2020-12-02
Add support for specifying maxPods as a bootstrap option

backjo:
This PR builds off of conversation in #184 . This support is initially only added for BottleRocket nodes. In a later PR, additional fields will be added, as well as support for these fields on any/all compatible OS families.


controllers/provisioners/eks/helpers.go
```diff
@@ -125,6 +129,9 @@ func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, k
 api-server   = "{{ .ApiEndpoint }}"
 cluster-certificate = "{{ .ClusterCA }}"
 cluster-name = "{{ .ClusterName }}"
+{{- if .MaxPods}}
```
eytan-avisror, line 132: Will this default to 0 if not provided? should we have a better default?

eytan-avisror:
I thought the agreement in #184 was only to add structure around bootstrap arguments vs. kubelet arguments which might be difficult to keep track of.

https://github.com/keikoproj/instance-manager/issues/184#issuecomment-720664630

Can you add a full design of how this should look like after we support all os families, what options we want to support and how the API will look like?

Since this adds a new API I want to make sure we are in full agreement on how it will look like.
Feel free to add it here or in #184

backjo:
> Another option that might be easier is to not exclude kubelet-extra-args and merge the two fields for now (similar to what you suggested), and deprecate `bootstrapArguments` field in v1alpha2, so that you can submit something like:
> 
> ```yaml
> configuration:
>   bootstrapArguments: "--foo=bar"
>   bootstrapOptions:
>     kubeletArguments: "--eviction-hard memory.available<300mi"
>     maxPods: 20 #jback changed this
>     dnsClusterIP: 10.0.0.1 #jback added this
>     dockerConfigJson: | 
>        {
>  	"bridge": "none",
>  	"log-driver": "json-file",
>  	"log-opts": {
>  		"max-file": "5",
>  		"max-size": "50"
>        }
>      }
> ```
> 
> Would result in:
> 
> ```
> bootstrap.sh <clustername> --use-max-pods 'false' --docker-config-json "<JSON payload>" --kubelet-extra-args '--max-pods "20" --eviction-hard "memory.available<300mi" --foo="bar" '
> ```

Going off of this ^ proposal (changed useMaxPods to maxPods). The example spec provided above would translate as follows:

AL2
```
bootstrap.sh <clustername> --dns-cluster-ip 10.0.0.1 --use-max-pods 'false' --docker-config-json "<JSON payload>" --kubelet-extra-args '--max-pods "20" --eviction-hard "memory.available<300mi" --foo="bar" '
```


BottleRocket
```
[settings.kubernetes]
api-server   = <autogenerated>
cluster-certificate = <autogenerated>
cluster-name = <clustername>
cluster-dns-ip = "10.0.0.1"
max-pods = 20
```

Windows
TODO - not documented so i need to go look at the PowerShell script on a real host machine. 

I'd advocate for sticking to this small surface area of properties, as they are all shared by at least 2 of the 3 OS families.

controllers/provisioners/eks/helpers.go
backjo: it will  default to 0 and be falsey

api/v1alpha1/instancegroup_types.go
```diff
@@ -179,6 +179,10 @@ type EKSManagedSpec struct {
 	EKSManagedConfiguration *EKSManagedConfiguration `json:"configuration"`
 }
 
+type BootstrapOptions struct {
+	MaxPods int64 `json:"maxPods,omitempty"`
```
eytan-avisror, line 183: are we planning to add a `KubeletArguments` element struct down the road to support all possible kubelet options in free form? if so how would we handle a `--max-pods` flag provided there directly?
e.g.
```
configuration:
  bootstrapOptions:
    kubeletArguments: "--max-pods 10 --eviction-hard memory.available<300mi"
    maxPods: 20
```
Since you are abstracting the UseMaxPods bool into just taking the maxPods (which is definitely nicer), we need to think of what to do in case of 'conflict' (this is only a problem in non-bottlerocket OS).

Also, what will we do in cases where we get both bootstrapArguments and `kubeletArguments` prior to deprecation of `bootstrapArguments`? e.g. 
```
configuration:
  bootstrapArguments: "--eviction-hard memory.available<600mi"
  bootstrapOptions:
    kubeletArguments: "--eviction-hard memory.available<300mi"
```

backjo: Great questions!

Given that we don't have an API for BottleRocket kubelet args at the moment, I don't want to get too deep into designing an API for it. I think a freeform string for now works - it's likely that for bottlerocket, we could leverage the same field to just pass in TOML. 

I'm thinking that we just allow duplicate flags. I ran into this while testing, actually - Kubelet just uses the last argument passed in. With that in mind, I think we should go the following strategy:

```bash
--kubelet-extra-args '<bootstrapArguments> <kubeletArguments> <generatedArguments'
```

This will result in:
- Generated kubelet args having the highest priority
- The 'new' kubeletArguments having normal priority
- The to-be-deprecated bootstrapArguments having the lower priority.

eytan-avisror: Awesome, this approach should be good !

eytan-avisror:
@backjo 
I'll merge this shortly - I think before we release 0.10.0 though we should complete this feature as discussed above with support for all os families as well as documentation that describes the behavior you mentioned above.
For the meanwhile, let me know if you need an image with this change and I will push one for you.
🥇

eytan-avisror (approved):

backjo:
Sounds good!

PR #221, opened 2021-01-04
Config exclusion

### LINKED ISSUE #219: Config exclusion option
We should allow users to exclude specific namespaces from using the global config / boundaries.

For example, if a namespace presents an annotation, instance groups under that namespace would avoid taking default values / boundary configuration from the configmap.

This is useful in case specific namespaces such as admin/system managed namespaces want an opt-out, while other tenant namespaces will continue to use it.

eytan-avisror:
Closes #219 

This PR adds the ability to exclude a namespace from global config, and improves bdd runtime by using this feature to run configmap tests in parallel in a different namespace.

- [x] BDD passing


controllers/reconcilers.go
```diff
@@ -127,6 +135,70 @@ func (r *InstanceGroupReconciler) configMapReconciler(obj handler.MapObject) []c
...
+	}
+
+	ns := obj.Object.(*corev1.Namespace)
+
+	r.NamespacesLock.Lock()
+	defer r.NamespacesLock.Unlock()
+
+	if _, ok := r.Namespaces[name]; !ok {
+		// new namespace
+		r.Namespaces[name] = *ns
```
kevdowney, line 168: So are you caching all the namespaces or just the ones w/ expected annotation?

eytan-avisror: Caching all namespaces, check for annotation happens downstream.
We already have a watch established by the controller, so it should be low significance in caching all of them

kevdowney: This has a code "smell" to me... looks like this caches entire object which granted is maybe not too large but it also doesn't remove it from the cache.

It would be better IMO to cache only the namespaces with the expected annotation which you can do here in this method as a filter to the cache, then also determine if you really need the entire object since your watch client is gonna already be caching the entire object for you. Lastly, you'd need to remove namespaces from the cache when a ns is deleted.

eytan-avisror: There is deletion if you look closely.
You make some good points, but the downside of making this specific to excluded namespaces and checking for the annotation in the reconciler is that if tomorrow we have a new annotation we will need to change the data structure again.

eytan-avisror: We could potentially keep the data structure same by still caching the entire object (map[string]corev1.Namespace) and just move the annotation check upstream to the reconciler, which would only cache namespaces with specific annotations, but we would still need to make the check downstream if in the future there are multiple namespace annotations we check for, so in that sense it doesnt really make sense to do it twice.
Alternatively we could have separate maps for separate annotations e.g. ExcludedConfigNamespaces, SomeOtherAnnotationNamespaces etc.
@kevdowney

controllers/reconcilers.go
```diff
@@ -127,6 +135,70 @@ func (r *InstanceGroupReconciler) configMapReconciler(obj handler.MapObject) []c
...
+		Name:      name,
+	}
+
+	r.NamespacesLock.Lock()
+	defer r.NamespacesLock.Unlock()
+
+	err := r.Get(context.Background(), namespacedName, obj.Object)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			r.Log.Info("namespace deleted", "object", name)
```
kevdowney, line 156: You should test this, if your watch only triggers on deletionTimestamp being set then your code here will likely never be triggered.

controllers/reconcilers.go
kevdowney: Since it's namespace object is not too much info, seems ok. Using exclusion maps though might be the right approach for long-term maintainability as it will not have to implement logic checks everywhere.

kevdowney (approved):

controllers/reconcilers.go
eytan-avisror: Tested this, it correctly detects deletion. a watch event is triggered on delete, and in that case the object is no longer found.

controllers/reconcilers.go
eytan-avisror: I think it would be more efficient to keep a single map and do logical checks in the controller vs. keep multiple maps which are specific to filters - scalability wise the more annotations we monitor the more maps we will have to maintain - with this approach we always maintain a single map regardless of number of annotations we monitor

PR #224, opened 2021-01-23
fix: remove explicit architecture in go build

backjo:
The explicit GOARCH variable in the Dockerfile causes ARM images to contain an AMD64 executable. This variable is unnecessary, since go will inspect the system to figure out what architecture it is running in. 


eytan-avisror (approved):
👍

PR #225, opened 2021-01-26
Avoid launch template creation loop

### LINKED ISSUE #223: excessing creation of LaunchTemplates if ScalingGroup creation fails.
**Is this a BUG REPORT or FEATURE REQUEST?**: Bug report

**What happened**: When creating an InstanceGroup with an incorrect AMI (one that doesn't exist for example), instance-manager logs the error but creates a new LaunchTemplate on each reconciliation loop.

**What you expected to happen**: Only one LaunchTemplate is created

**How to reproduce it (as minimally and precisely as possible)**: Create an InstanceGroup using LaunchTemplates with an non-existent AMI ID .

**Anything else we need to know?**: AWS incorrectly reports this error as `You are not authorized to use launch template` see https://coding-stream-of-consciousness.com/2019/12/02/aws-not-authorized-to-use-launch-template-in-terraform-or-in-console/ for some examples.

**Environment**:

- Kubernetes version:
clientVersion:
  buildDate: "2020-09-16T21:51:49Z"
  compiler: gc
  gitCommit: f5743093fd1c663cb0cbc89748f730662345d44d
  gitTreeState: clean
  gitVersion: v1.19.2
  goVersion: go1.15.2
  major: "1"
  minor: "19"
  platform: darwin/amd64
serverVersion:
  buildDate: "2020-10-20T23:27:12Z"
  compiler: gc
  gitCommit: ad4801fd44fe0f125c8d13f1b1d4827e8884476d
  gitTreeState: clean
  gitVersion: v1.16.15-eks-ad4801
  goVersion: go1.13.15
  major: "1"
  minor: 16+
  platform: linux/amd64

```bash
kubectl version -o yaml
```

eytan-avisror:

Fixes #223 

Background: This bug only happens on the creation flow - when the launch template is created, but the scaling group fails to create due to a launch template problem (e.g. bad AMI, missing permission, etc). In this case we never move on to update flow since the scaling group never gets created, and there is no logic to handle the case where the launch template already exist but the scaling group does not - so we enter an endless loop of creating new configurations until the CR value is fixed.

Desired behavior: The behavior should be that we create a scaling config once, if the downstream operation (e.g. creating a scaling group) fails due to a bad config, we should only create a new config/config version if there is a diff with the one we already created.
So if the user creates a resource with a bad AMI, we create a single config, and when we fail to use it, we continue failing without creation of new configs or config versions.
Once the resource has changed to reflect a different AMI (or other config), we create a new config version and try to use that instead.

- Discover launch templates on create flow as well - pass in a specific config name `TargetConfigName` in case it's cached in status in order to detect created launch config / template even if scaling group creation has failed in a previous reconcile.

- Always call `scalingConfig.Create()` in create scenario, the internal logic of `Create()` already checks if `IsProvisioned()` or not.

- Check if there is a drift before creating a new launch template version - if there is no drift avoid creating endless number of launch template versions until the problem is corrected via the CR.

- [x] BDD Passing
- [x] Test Switching between Template/Config

### DIFF
--- a/api/v1alpha1/instancegroup_types.go
+++ b/api/v1alpha1/instancegroup_types.go
@@ -914,10 +914,13 @@ func (status *InstanceGroupStatus) GetActiveLaunchTemplateName() string {
 
 func (status *InstanceGroupStatus) SetActiveLaunchConfigurationName(name string) {
 	status.ActiveLaunchConfigurationName = name
+	status.ActiveLaunchTemplateName = ""
+	status.LatestTemplateVersion = ""
 }
 
 func (status *InstanceGroupStatus) SetActiveLaunchTemplateName(name string) {
 	status.ActiveLaunchTemplateName = name
+	status.ActiveLaunchConfigurationName = ""
 }
 
 func (status *InstanceGroupStatus) SetLatestTemplateVersion(version string) {
--- a/controllers/provisioners/eks/cloud.go
+++ b/controllers/provisioners/eks/cloud.go
@@ -78,15 +78,40 @@ func (ctx *EksInstanceGroupContext) CloudDiscovery() error {
 	status.SetLifecycle(v1alpha1.LifecycleStateNormal)
 
 	if spec.IsLaunchConfiguration() {
-		state.ScalingConfiguration = &scaling.LaunchConfiguration{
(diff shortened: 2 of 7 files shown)


tekenstam (approved):
LGTM

PR #233, opened 2021-02-03
feat: add support for new volume types

backjo:

### DIFF
--- a/api/v1alpha1/instancegroup_types.go
+++ b/api/v1alpha1/instancegroup_types.go
@@ -111,12 +111,6 @@ var (
 	LifecycleHookAllowedDefaultResult   = []string{LifecycleHookResultAbandon, LifecycleHookResultContinue}
 	LaunchTemplatePlacementTenancyTypes = []string{HostPlacementTenancyType, DefaultPlacementTenancyType, DedicatedPlacementTenancyType}
 	log                                 = ctrl.Log.WithName("v1alpha1")
-
-	volumeTypesWithProvisionedIopsSupport = map[string]bool{
-		"io1": true,
-		"io2": true,
-		"gp3": true,
-	}
 )
 
 // InstanceGroup is the Schema for the instancegroups API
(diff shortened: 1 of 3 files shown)


api/v1alpha1/instancegroup_types.go
```diff
@@ -111,6 +111,12 @@ var (
 	LifecycleHookAllowedDefaultResult   = []string{LifecycleHookResultAbandon, LifecycleHookResultContinue}
 	LaunchTemplatePlacementTenancyTypes = []string{HostPlacementTenancyType, DefaultPlacementTenancyType, DedicatedPlacementTenancyType}
 	log                                 = ctrl.Log.WithName("v1alpha1")
+
+	volumeTypesWithProvisionedIopsSupport = map[string]bool{
```
eytan-avisror, line 115: minor nit:
Should we move this to awsprovider along with `AllowedVolumeTypes`?
We might want a shorter name too - how about `AllowedVolumeTypesWithProvisionedIOPS`?

Also what are you thoughts on using a string slice (such as `AllowedVolumeTypes`) vs. a map like you implemented here?

We can then remove the map check and change the if condition to something like:
```yaml
if v.Iops != 0 && !common.ContainsEqualFold(awsprovider.AllowedVolumeTypesWithProvisionedIOPS, v.Type) {
  ...
}
```

This is minor and both options are good, but just might be more in line with existing code.

eytan-avisror:
LG, added some minor suggestion to refactor.
Assume you verified manually this works? 🥇

api/v1alpha1/instancegroup_types.go
backjo: Sounds good - will make this check!

backjo:
Yep, verified this works on our EKS clusters. One caveat that might be applicable for some folks is that the LaunchConfiguration APIs don't support the new volume types - only the LaunchTemplate APIs

eytan-avisror:
> Yep, verified this works on our EKS clusters. One caveat that might be applicable for some folks is that the LaunchConfiguration APIs don't support the new volume types - only the LaunchTemplate APIs

In that case, should we add another validation above the provisioned IOPS one to validate if user is trying to use invalid types with template/config?

We can add `TemplateAllowedVolumeTypes` & `ConfigurationAllowedVolumesTypes`
You can use predicates `spec.IsLaunchTemplate()` or `spec.IsLaunchConfiguration()` to check what the CR is declaring.

backjo:
Added validation based on ScalingConfigurationType

eytan-avisror (approved):
LG
Thanks!

PR #236, opened 2021-02-09
remove config hash when namespace is excluded

### LINKED ISSUE #235: Infinite reconcile loop in Kubernetes 1.18
Seems like we get stuck in an infinite reconcile loop when running on Kubernetes 1.18.
It seems that even though we are using Status subresource / controller-runtime StatusWriter to modify status only, even though there are no changes, a reconcile is still triggered.

The result is a never ending reconcile.

eytan-avisror:

Fixes #235 

- Moves setting of config hash up to controller level, will now unset configHash field if configmap is nil/empty/does not exist.
- Refactors namespace exclusion logic into IsNamespaceExcluded method
- Refactors `UpdateStatus` to use patch instead of update, in order avoid metadata changes causing a reconcile loop - in k8s 1.18 managedFields was introduced along with a `time` field which causes this problem when using Update method. For now this implements a custom StatusPatch, in the future we might want something similar to #240 

- [x] BDD Passing

@backjo

### DIFF
--- a/controllers/instancegroup_controller.go
+++ b/controllers/instancegroup_controller.go
@@ -119,7 +119,7 @@ func (r *InstanceGroupReconciler) Reconcile(ctxt context.Context, req ctrl.Reque
 		r.Log.Error(err, "reconcile failed")
 		return ctrl.Result{}, err
 	}
-	existingStatus := instanceGroup.Status
+	statusPatch := kubeprovider.MergePatch(*instanceGroup)
 
 	// set/unset finalizer
 	r.SetFinalizer(instanceGroup)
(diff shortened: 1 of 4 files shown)


controllers/instancegroup_controller.go
```diff
@@ -184,34 +179,61 @@ func (r *InstanceGroupReconciler) Reconcile(ctxt context.Context, req ctrl.Reque
...
 		if kubeprovider.IsStorageError(err) {
 			return
 		}
-		r.Log.Info("failed to update status", "error", err, "instancegroup", ig.NamespacedName())
+		r.Log.Info("failed to update status", "error", err, "instancegroup", instanceGroup.NamespacedName())
+	}
+}
+
+func (r *InstanceGroupReconciler) IsNamespaceExcluded(instanceGroup *v1alpha1.InstanceGroup) bool {
+	if reflect.DeepEqual(r.ConfigMap, &corev1.ConfigMap{}) {
```
backjo, line 220: Is this check necessary? Seems like it's outside of the scope of what this function should need to do (and, in fact, is checked already before calling this function). Semantically, I'd assume that IsNamespaceExcluded is a separate check from whether or not there is actually a managed configuration

controllers/instancegroup_controller.go
```diff
@@ -184,34 +179,61 @@ func (r *InstanceGroupReconciler) Reconcile(ctxt context.Context, req ctrl.Reque
...
 		// avoid error if object already deleted
 		if kubeprovider.IsStorageError(err) {
 			return
 		}
-		r.Log.Info("failed to update status", "error", err, "instancegroup", ig.NamespacedName())
+		r.Log.Info("failed to update status", "error", err, "instancegroup", instanceGroup.NamespacedName())
+	}
+}
+
+func (r *InstanceGroupReconciler) IsNamespaceExcluded(instanceGroup *v1alpha1.InstanceGroup) bool {
```
backjo, line 219: Maybe name this IsNamespaceExcludedFromManagedConfiguration to be more explicit about what it's excluded from?

eytan-avisror: I'll make this more generic by taking in a key value and rename to `IsNamespaceAnnotated`

controllers/instancegroup_controller.go
eytan-avisror: Yep, this is probably redundant

controllers/instancegroup_controller.go
backjo: 👍

eytan-avisror:
```
27 scenarios (27 passed)
147 steps (147 passed)
28m47.254921999s
PASS
ok  	github.com/keikoproj/instance-manager/test-bdd	1727.554s [no tests to run]
```

backjo (approved):

PR #237, opened 2021-02-09
fix: set latest launch template version when creating a new version

### LINKED ISSUE #239: LaunchTemplateVersion is not set after updating the scaling group
Seems we only set LaunchTemplateVersion status field during Discover, and not directly after we update the scaling group.
This causes the upgrade CR to also have the old version as unique identifier instead of the newer one.

backjo:

Fixes #239

### DIFF
--- a/controllers/providers/aws/aws.go
+++ b/controllers/providers/aws/aws.go
@@ -189,13 +189,12 @@ func (w *AwsWorker) UpdateLaunchTemplateDefaultVersion(name, defaultVersion stri
 	return out.LaunchTemplate, nil
 }
 
-func (w *AwsWorker) CreateLaunchTemplateVersion(input *ec2.CreateLaunchTemplateVersionInput) (int64, error) {
-	var version int64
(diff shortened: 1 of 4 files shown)


eytan-avisror:
@backjo 
Looks good 🥇 
BDD passing on this?

backjo:
Haven't run it yet - need to spin up an environment where it can run :)

eytan-avisror:
> Haven't run it yet - need to spin up an environment where it can run :)

No worries, I can run it soon

backjo:
> > Haven't run it yet - need to spin up an environment where it can run :)
> 
> No worries, I can run it soon

Thanks!

eytan-avisror:
BDDs are passing, just doing more testing around concurrency policies to make they work as expected with these changes

eytan-avisror (approved):
👍 
There is a problem with Allow concurrency policy w/ LaunchTemplates, but was probably there before this PR - will open a separate bug to address it.

PR #244, opened 2021-02-10
feature: add support for calculating custom networking max pods

backjo:
This PR adds support for automatic calculation of the max pods bootstrap argument. This is useful for EKS clusters where custom networking is enabled, because the default maxPods value is not correct. See https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html for more information about this limitation. 


controllers/provisioners/eks/helpers.go
```diff
@@ -457,14 +458,46 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
...
 		instanceGroup = ctx.GetInstanceGroup()
 		configuration = instanceGroup.GetEKSConfiguration()
 	)
+	if instanceGroup.GetAnnotations()[CustomNetworkingEnabledAnnotation] == "true" {
+		hostNetworkPods, err := strconv.ParseInt(instanceGroup.GetAnnotations()[CustomNetworkingHostPodsAnnotation], 10, 64)
+		if err != nil {
+			hostNetworkPods = 2 //Default on EKS. Kube-Proxy and AWS VPC CNI
+		}
+
+		instanceTypes, _ := ctx.AwsWorker.DescribeInstanceTypes()
```
eytan-avisror, line 472: Can you access the cache instead of making another call?

You might have to make the actual call not only in the case of launch template here:
https://github.com/keikoproj/instance-manager/blob/master/controllers/provisioners/eks/cloud.go#L233-L235
You can then cache it in DiscoveredState and use it here.

We try to avoid making additional read calls within the reconcile and rely on the state discovered in the beginning of the reconcile.

backjo: Yep, makes sense

controllers/provisioners/eks/helpers.go
```diff
@@ -457,14 +458,47 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
...
 		configuration = instanceGroup.GetEKSConfiguration()
 	)
+	if instanceGroup.GetAnnotations()[CustomNetworkingEnabledAnnotation] == "true" {
+		hostNetworkPods, err := strconv.ParseInt(instanceGroup.GetAnnotations()[CustomNetworkingHostPodsAnnotation], 10, 64)
+		if err != nil {
+			hostNetworkPods = 2 //Default on EKS. Kube-Proxy and AWS VPC CNI
+		}
+
+		instanceTypes := state.GetInstanceTypeInfo()
+		for _, instanceType := range instanceTypes {
```
eytan-avisror, line 474: Wonder if you can simplify this logic a little bit by adding a utility method under awsprovider e.g.

```go
// This will remove loop & first if
networkInfo := awsprovider.GetInstanceTypeNetworkInfo(types, type)
maxPods := (networkInfo.MaximumNetworkInterfaces-1) * (networkInfo.Ipv4AddressesPerInterface-1) + hostNetworkPods
if configuration.BootstrapOptions == nil {
        // Can avoid else if there it returns
	return &v1alpha1.BootstrapOptions{
		MaxPods: maxPods,
	}
}
if bootstrapOptions.MaxPods == 0 {
	bootstrapOptions.MaxPods = maxPods
}
return &bootstrapOptions
```

WDYT?

controllers/provisioners/eks/cloud.go
```diff
@@ -205,6 +206,12 @@ func (ctx *EksInstanceGroupContext) CloudDiscovery() error {
 	status.SetCurrentMin(int(aws.Int64Value(targetScalingGroup.MinSize)))
 	status.SetCurrentMax(int(aws.Int64Value(targetScalingGroup.MaxSize)))
 
+	instanceTypes, err := ctx.AwsWorker.DescribeInstanceTypes()
+	if err != nil {
+		return errors.Wrap(err, "failed to discover launch templates")
```
eytan-avisror, line 211: Probably want to tweak this error message

controllers/provisioners/eks/helpers.go
backjo: Yeah, totally agree !

eytan-avisror (approved):
Great feature !

PR #249, opened 2021-02-11
feature: Add MaxRetries to CRD Strategy

eytan-avisror:

This PR continues work done in #243 
Since deletion of inactive resources is now removed, behavior has changed so that if the upgrade fails it does not retry and enters an error state - previous to this change, the upgrade would be retried to infinity (or until success).

This PR adds a MaxRetries field to CRD Strategy:
```
spec:
  provisioner: eks
  strategy:
    type: crd
    crd:
      maxRetries: 2
```

By default, if not set, maxRetries is set to `3`.
Controller will cache current retry count in IG status, and reset to 0 when:
- Rotation no longer needed (all nodes have rotated)
- New custom resource is submitted (e.g. the failed / timed out resource is manually deleted)
- Resource is completed successfully


backjo (approved):

controllers/providers/kubernetes/crd.go
```diff
@@ -170,11 +172,33 @@ func ProcessCRDStrategy(kube dynamic.Interface, instanceGroup *v1alpha1.Instance
 
 	if strings.EqualFold(resourceStatus, successString) {
 		log.Info("custom resource succeeded", "instancegroup", instanceGroupNamespacedName, "resource", customResourceName, "status", resourceStatus)
+		status.SetStrategyRetryCount(0)
 		return true, nil
 	}
 
 	if strings.EqualFold(resourceStatus, failureString) {
 		log.Info("custom resource failed", "instancegroup", instanceGroupNamespacedName, "resource", customResourceName, "status", resourceStatus)
+		maxRetries := *strategy.MaxRetries
```
shrinandj, line 181: One other option is to keep decrementing the strategy.MaxRetries until it reaches 0. Avoids using an extra variable...

PR #252, opened 2021-02-11
fix: upgrade resource replacement loop

### LINKED ISSUE #251: Upgrade keeps replacing resource
Looks like we messed up the iteration back in #179 where we did:
```
for _, r := range r.Items {
```

There are also other issues leading to activeResource detection having the wrong resources in the list when many instance groups are upgrading simultaneously - the end result of this is that we keep deleting the upgrade resource and replacing it with a new one.

In addition, there is a problem when we go to upgrade straight after updating the ASG with a new launch config - since the cache still has a reference to old ASG with old launch config - this causes recreation of the upgrade resource after the initial submission. - we should requeue right after / if we make an update to the scaling group.

eytan-avisror:

Fixes #251 

- [x] BDD passing
- [x] Manual testing of all policies w/ template/config

### DIFF
--- a/controllers/provisioners/eks/update.go
+++ b/controllers/provisioners/eks/update.go
@@ -275,6 +275,7 @@ func (ctx *EksInstanceGroupContext) ScalingGroupUpdateNeeded(configName string) 
 		}
 	case scalingGroup.MixedInstancesPolicy != nil:
 		name = aws.StringValue(scalingGroup.MixedInstancesPolicy.LaunchTemplate.LaunchTemplateSpecification.LaunchTemplateName)
+		scalingGroup.MixedInstancesPolicy.LaunchTemplate.LaunchTemplateSpecification.LaunchTemplateId = nil
 		if desiredPolicy == nil {
 			return true
 		}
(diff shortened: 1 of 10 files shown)


controllers/provisioners/eks/update.go
```diff
@@ -111,6 +111,11 @@ func (ctx *EksInstanceGroupContext) Update() error {
 		return errors.Wrap(err, "failed to update scaling group")
 	}
 
+	if updated {
+		// requeue after scaling group update occurs to refresh cache
```
backjo, line 115: Is this something we want to be durable long-term, or just for now to fix the loop? Feels like we should update cache eventually rather than relying on the next reconcile to do it.

backjo (approved):

controllers/provisioners/eks/update.go
eytan-avisror: I find from past experience with this & other controllers it's much easier to manage the state using this pattern of having short reconciles and requeue when you need to, otherwise you either find yourself making excessive amount of calls or  working on semi-accurate data.

Excessive amount of API calls is not that bad actually now that we use https://github.com/keikoproj/aws-sdk-go-cache 
we used to get throttled heavily before on clusters with many InstanceGroups/other.

Issue #197, opened 2020-11-14
rollingUpdate strategy - draining, and additional options


eytan-avisror:
Currently, the rollingUpdate strategy implementation is very basic, we should make some basic improvements to it so that it is more useful.

- Add node draining
- Configurable waits

PR #259, opened 2021-02-17
(WIP) feature: Drain nodes prior to termination when using RollingUpgrade strategy

eytan-avisror:
Adds to #197 

This feature adds a capability to do draining in parallel via kubectl as a library (not shelling out), prior to termination of instances.

This makes the rollingUpgrade strategy much more usable.

Implementation adds a sync.Map of `namespacedName` : `sync.WaitGroup` at the controller level in order to keep track of draining nodes.

The drain is initiated and then the controller will re-queue until operation has completed or errors out.
Also refactored to move the rollingUpgrade logic from kubeprovider into the eks package.

TBD:
- [ ] Write unit tests
- [ ] Test edge cases (PDB, drain error, autoscaler, etc)


controllers/provisioners/eks/rollingupdate.go
```diff
@@ -0,0 +1,301 @@
...
+	drainErrs := make(chan error)
+	if reflect.DeepEqual(r.DrainGroup, &sync.WaitGroup{}) {
+		for _, node := range targetNodes.Items {
+
+			nodeName := node.GetName()
+			r.Info("creating drainer goroutine for node", "node", nodeName)
+
+			drainOpts := &drain.Helper{
+				DeleteLocalData:     true,
+				Force:               true,
```
backjo, line 246: Might consider making force (and others properties?) configurable (maybe at the IG level?). I could see a use case (say ephemeral pods for build agents on something like Jenkins) where an operator might want to not force the drain, but rather retry until ephemeral pods terminate naturally.

eytan-avisror: Yes, I wanted to expose only Timeout, but can add Force as well.
I will also add some basic implementation of readiness gates

```yaml
spec:
  strategy:
    type: rollingUpdate
    rollingUpdate:
      maxUnavailable: 2
      readinessGates:
      # All pods in the namespace must be ready in order to proceed
      - type: Namespace
        value: namespace-1
      drainOptions:
        timeoutSeconds: 900
        force: true
```

One thing that is a bit challenging with this implementation is how to properly implement timeout for the batch of rotating nodes - since we requeue instead of blocking the thread while waiting for drain - might need to persist a timestamp in status, but that doesnt feel like the best solution.

PR #265, opened 2021-03-31
fix: Resolve caching leak

### LINKED ISSUE #264: Memory leak in pagination of DescribeInstanceTypes
We make the call `DescribeInstanceTypes` on every reconcile, but cache the result for up to 24hrs.
The payload of this call is very large, and usually comes in 3-4 full pages and hardly ever changes over time (when new instance types are added, or existing instance types are changed).

It seems our caching library `aws-sdk-go-cache` is leaking the paginated calls, and in this case causing a memory leak (because the call will always paginate and is massive payload).

The caching library creates a cache per operation, e.g. ec2.DescribeInstanceTypes is it's own cache, and stores the data with `request.Params` as key, and `response.Body` as value, meaning that a call with input X will always have output Y.

Since in the case for pagination, subsequent calls reference `NextToken` to access next pages, and `NextToken` is always a random string, this means we keep leaking these pages in the cache until the pod eventually gets OOMKilled.

Fix for pagination in the caching library is not likely since we cannot uniquify the parameters in the case for pagination.

This leaves following options:
- Check in a file with the payload - will have to re-validate every release whether there are changes.
- Make the call on controller load time - this will force all users to make this call regardless of whether they are using the eks provisioner or other, which may be unnecessary
- Find a way to avoid pagination when making the call by using filters
- Come up with an innovative fix in the https://github.com/keikoproj/aws-sdk-go-cache for the pagination problem

Either way, we should also reduce the cache MaxSize to avoid blowing up the memory, currently its set to 5000 items and prune 500 when it reaches max, we should probably go with a much smaller number e.g. 500/50 or even 300/30 since this MaxSize number is per operation cache and due to existing TTLs, these numbers should be appropriate (should test).

Screen Shot 2021-03-29 at 2 06 17 PM
Screen Shot 2021-03-29 at 4 34 10 PM

eytan-avisror:
Fixes #264 

This adds background pruning in cache based on TTL every hour, as well as reduce individual cache sizes to a more reasonable number e.g. 250 items per operation, and GC 25 when reach max 


backjo (approved):

eytan-avisror:
Confirmed this works properly with the caching fix.
Did a load test with the TTL set to 3 minutes and pruning every 15 (in order to see results quickly), one image with the caching fix vs one image without.

Screen Shot 2021-04-01 at 2 56 08 PM

PR #273, opened 2021-04-15
feat: conditional default values

backjo:


controllers/provisioners/config.go
```diff
@@ -151,14 +179,31 @@ func isConflict(defaultVal, resourceVal interface{}) bool {
 	return false
 }
 
-func (c *ProvisionerConfiguration) setSharedFields(obj map[string]interface{}) error {
+func getMatchingConditionals(ig *v1alpha1.InstanceGroup, conditionals []Conditional) []Conditional {
+	var applicableConditionals = make([]Conditional, 0)
+	for _, conditional := range conditionals {
+		annotationKeyValue := strings.Split(conditional.Annotation, "=")
+		if ig.Annotations[annotationKeyValue[0]] == annotationKeyValue[1] {
```
eytan-avisror, line 186: You can actually use providers/kubernetes/utils.go `HasAnnotation(ig.Annotations, annotationKeyValue[0], annotationKeyValue[1])`, might also want to pre validate the input `conditional.Annotation` is not empty and contains '=' sign to avoid crashes in case of bad input

controllers/provisioners/config_test.go
```diff
@@ -160,6 +169,333 @@ spec:
...
+    - spec.strategy`
+
+	mockConditionals := `
+- annotation: "instancemgr.keikoproj.io/os-family=windows"
+  defaults:
+    spec:
+      eks:
+        configuration:
+          image: ami-22222222222
+- annotation: "instancemgr.keikoproj.io/arch=arm64"
```
eytan-avisror, line 196: So if we have two annotations referencing the same default value, the last one always takes priority? how do we choose between the two?

Also should use the yaml content for defaults be inline since its unstructured?
```
- annotation: <annotation>
  defaults: |
    < inline yaml >
```

eytan-avisror:
Very cool. had some questions when you get the chance.. I may need to think about the edge cases of conflicts in default between multiple matching condtionals

controllers/provisioners/config_test.go
backjo: Yeah, the priority is just the last applicable rule that applies. We could consider doing something around a `priority` field where folks can determine explicitly what the ordering is.

backjo: w.r.t. the inline YAML - I don't think it's needed for defaults since the entire conditionals list is already defined like:
```
kind: ConfigMap
data: 
    conditionals: |
      - annotation: "x=y"
        defaults: 
          foo: bar
```

eytan-avisror: Is ordering kept though when we unmarshal to yaml? or do we have to sort? priority might be good.. maybe instead we should allow chaining of annotations:

```
- annotation: "instancemgr.keikoproj.io/os-family=windows,instancemgr.keikoproj.io/arch=arm64"
  defaults:
    spec:
      eks:
        configuration:
          image: ami-22222222222
```

would this be a better mechanism?

backjo: It seems to be preserved in the tests that I've run - and I'd expect it to remain ordered based on previous use - I think order issues tend to happen when folks want ordering on things that are not supposed to be ordered - typically k/v maps.

eytan-avisror: Maybe we can keep the current ordering, but still allow having conditionals with multiple annotations? although if there are multiple the string might get pretty long.. but it might be useful since it will allow avoiding thinking of order when there are multiple conditions for a default

backjo: Also might be worth looking into supporting standard selector syntax

eytan-avisror: This actually might be nicer and allow the chaining, and will also avoid the string parsing / validation. we can add more matchers later
```
selector:
  matchAnnotations:
    instancemgr.keikoproj.io/os-family: windows
    instancemgr.keikoproj.io/arch=arm64
  defaults:
    spec:
      eks:
        configuration:
          image: ami-22222222222
```

backjo: check out the latest commit - I ended up leveraging the K8S field selector library so we can do things like:

```yaml
- annotation: instancemgr.keikoproj.io/os-family = 'windows', instancemgr.keikoproj.io/arch: arm64
  defaults:
  spec:
    eks:
      configuration:
        image: ami-22222222222
- annotation: instancemgr.keikoproj.io/os-family in (bottlerocket, amazonlinux2)
  defaults:
  spec:
    eks:
      configuration:
        image: ami-333333333

```

docs/EKS.md
```diff
@@ -545,6 +545,51 @@ This also makes upgrades easier across a managed cluster, an operator can now si
...
+    - annotation: 'instancemgr.keikoproj.io/arch = arm64,instancemgr.keikoproj.io/os-family = bottlerocket'
+      defaults:
+       spec:
+         eks:
+           configuration:
+             image: ami-bottlerocketArm
+```
+
+The `annotation` field accepts a selector string, which follows the semantics of the Kubernetes [field selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/). 
+This selector is applied against the annotations of InstanceGroup objects. When a selector matches an InstanceGroup, the `defaults` associated with it are applied to the IG, taking precedence over the non-conditional
```
eytan-avisror, line 591: Can we change "IG" to instance group?

docs/EKS.md
```diff
@@ -545,6 +545,51 @@ This also makes upgrades easier across a managed cluster, an operator can now si
...
+             image: ami-windows
+    - annotation: 'instancemgr.keikoproj.io/arch = arm64,instancemgr.keikoproj.io/os-family = bottlerocket'
+      defaults:
+       spec:
+         eks:
+           configuration:
+             image: ami-bottlerocketArm
+```
+
+The `annotation` field accepts a selector string, which follows the semantics of the Kubernetes [field selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/).
```
eytan-avisror, line 590: We might want to provide a table with full syntax options, or a link that provides it - the above just mentions a few e.g. 

```
<selector-syntax>         ::= <requirement> | <requirement> "," <selector-syntax>
<requirement>             ::= [!] KEY [ <set-based-restriction> | <exact-match-restriction> ]
<set-based-restriction>   ::= "" | <inclusion-exclusion> <value-set>
<inclusion-exclusion>     ::= <inclusion> | <exclusion>
<exclusion>               ::= "notin"
<inclusion>               ::= "in"
<value-set>               ::= "(" <values> ")"
<values>                  ::= VALUE | VALUE "," <values>
<exact-match-restriction> ::= ["="|"=="|"!="] VALUE
```

controllers/provisioners/config_test.go
eytan-avisror: @backjo this is awesome!
I think the only downside, is that if someone wants to match multiple annotations it can become a super long string - but this seems super useful and we didn't have to implement any of our own parsing which is great.

eytan-avisror: Wonder if we should change the key "annotation" to "annotationSelector" WDYT?

backjo: yep, totally.

eytan-avisror (approved):
🥇

PR #274, opened 2021-04-16
feat: prometheus metrics integration

### LINKED ISSUE #238: Prometheus Metrics Integration
We should expose a prometheus endpoint and provide metrics.

- Instance Group per State - gauge that shows instance groups and their state
- Failed Reconcile - counter that shows number of failed reconcile events
- Successful Reconcile - counter that shows number of successful reconcile events
- Node Rotation - show last instance group upgrade per instance group
- AWS Throttling - counter of throttled calls
- AWS Cache - show cache miss/hits

eytan-avisror:
Fixes #238 

This adds a basic prometheus collector integration to expose metrics.


eytan-avisror:
Some examples on dashing the added metrics
Screen Shot 2021-04-16 at 1 30 14 PM
Screen Shot 2021-04-16 at 1 30 30 PM

controllers/common/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (c MetricsCollector) Describe(ch chan<- *prometheus.Desc) {
+	c.successCounter.Describe(ch)
+	c.failureCounter.Describe(ch)
+	c.throttleCounter.Describe(ch)
+	c.statusGauge.Describe(ch)
+}
+
+func (c *MetricsCollector) SetInstanceGroup(instanceGroup, state string) {
+	externalStates := []string{"ReconcileModifying", "InitUpgrade", "Deleting", "Ready", "Error"}
```
kevdowney, line 73: Are these states constants?

eytan-avisror: Yes, in v1alpha1 - this would be a cyclic dependency if we use them though

shaoxt (approved):
LGTM

PR #281, opened 2021-04-27
fix: run build on unit test task

backjo:

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -15,14 +15,18 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
-        go-version: ^1.13
+        go-version: ^1.15
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test.yml
```diff
@@ -15,11 +15,15 @@ jobs:
...
       with:
-        go-version: ^1.13
+        go-version: ^1.15
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Build
+      run: |
+        make go
```
eytan-avisror, line 25: you can use `make manager` for a simple build + vet

.github/workflows/unit-test.yml
```diff
@@ -15,11 +15,15 @@ jobs:
...
+
     - name: Test
       run: |
         go test -v ./controllers/... -coverprofile coverage.txt
```
eytan-avisror, line 29: we might want to change this to `make test` while we are doing this

eytan-avisror (approved):
👍

PR #284, opened 2021-04-29
fix: only set restricted field to default value if none of the conditionals set it

### LINKED ISSUE #283: Conditional restricted fields are not applied if the last applicable rule did not set the field.
BUG
**What happened**:
When there are multiple conditional rules, fields are set from the default rule if the field was not set in the last conditional rule.
 
```yaml
mockBoundaries := `
    restricted:
    - spec.eks.configuration.keyPairName
    - spec.eks.configuration.taints
    - spec.eks.configuration.labels
    - spec.eks.configuration.securityGroups
    - spec.eks.configuration.instanceType
    - spec.eks.configuration.image
    - spec.strategy`

mockConditionals := `
- annotationSelector: "instancemgr.keikoproj.io/os-family=windows"
  defaults:
    spec:
      eks:
        configuration:
          image: ami-22222222222
- annotationSelector: "instancemgr.keikoproj.io/arch in (arm64), instancemgr.keikoproj.io/os-family in (windows)"
  defaults:
    spec:
      eks:
        configuration:
          image: ami-33333333333
- annotationSelector: "instancemgr.keikoproj.io/arch in (arm64), instancemgr.keikoproj.io/os-family in (windows)"
  defaults:
    spec:
      eks:
        configuration:
          foo: bar

mockDefaults := `
spec:
  strategy:
    type: rollingUpdate
    rollingUpdate:
      maxUnavailable: 30%
  eks:
    configuration:
      keyPairName: TestKeyPair
      image: ami-025bf02d663404bbc
      securityGroups:
      - sg-123456789012
      instanceType: m5.large
      labels:
        label-key: label-value
      taints:
      - key: taint-key
        value: taint-value
        effect: NoSchedule`
```          
With the following setup, a IG that passes the `instancemgr.keikoproj.io/arch in (arm64), instancemgr.keikoproj.io/os-family in (windows)` selector would have the base image, ami-025bf02d663404bbc, applied - which is not expected. It is expected in this case that ami-33333333333 would be applied.

backjo:
fixes #283 

### DIFF
--- a/controllers/provisioners/config.go
+++ b/controllers/provisioners/config.go
@@ -168,17 +168,17 @@ func (c *ProvisionerConfiguration) setRestrictedFields(unstructuredInstanceGroup
 	for _, pathStr := range c.Boundaries.Restricted {
 		path := common.FieldPath(pathStr)
 		// if a default value exists for the path, set it on the instance group
-		var setField = false
-		var field interface{}
+		var setFieldInAnyConditional = false
 		for _, conditional := range applicableConditionals {
-			if field, setField, _ = unstructured.NestedFieldCopy(conditional.Defaults, path...); setField {
+			if field, setField, _ := unstructured.NestedFieldCopy(conditional.Defaults, path...); setField {
+				setFieldInAnyConditional = true
 				err := unstructured.SetNestedField(unstructuredInstanceGroup, field, path...)
 				if err != nil {
 					errors.Wrap(err, "failed to set nested field")
 				}
 			}
 		}
-		if setField {
+		if setFieldInAnyConditional {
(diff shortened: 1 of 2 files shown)


eytan-avisror:
So this would set all conditional with the first match?
Would we need a hotfix btw?

backjo:
Worse, actually. It would only set fields that were in the last match, since setField was getting reset for every applicable conditional. If the field wasn't set by the last conditional, it would override with the default value, if one existed.

backjo:
Yeah, a patch release would probably make sense.

eytan-avisror (approved):

PR #285, opened 2021-04-29
release 0.11.1

backjo:

### Fixed

- Only set restricted field to default value if none of the conditionals set it (#284)


eytan-avisror (approved):
👍

Issue #277, opened 2021-04-20
Support warm pools


eytan-avisror:
ASGs support provisioning pre warmed instances to make scaling faster.
This could be pretty useful if it cuts down the autoscaling time.

We should explore both running/stopped option for pre-warmed instances and support this API if it makes sense

https://docs.aws.amazon.com/sdk-for-go/api/service/autoscaling/#AutoScaling.PutWarmPool

- [x] AL2 Support
- [x] Windows Support
- [ ] Bottlerocket Support

backjo:
It'd be interesting to compare this to the existing startup time - it's around 90-120 seconds for us on our BottleRocket nodes.

Though, it looks like it might not work right now for us:
`Warm pools currently can't be used with ECS, EKS, and self-managed Kubernetes.`

eytan-avisror:
I just tested it in stopped mode, and it works! got nodes to join within few seconds of scale event.. not sure why they say it's not supported 🤣  need to see exactly what is not supported

eytan-avisror:
Pre warmed stopped instance - 1m 20s from pod scale event, to node ready
No pre-warming - 1m 50s - so it potentially shaves a cool 30s off of your scaling time
so around ~30% faster 

Not much, but nice improvement if you have very peaky traffic

eytan-avisror:
One corner case we need to work around somehow, is that when the warm pool instances are started they briefly join the cluster and then when they are stopped they leave.
The problem cases I've seen is when:
- They go to ready for some time (this may cause scheduling)
- The node objects sometime stick around as NotReady after the instance is stopped

Ideal solution would have a way to not run the bootstrap script on the prewarmed instances, and only run it when scaling happens.

I am experimenting with using some simple script addition in userdata to be able to do that e.g.
```
INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id)
LIFECYCLE=$(aws autoscaling describe-auto-scaling-instances --region us-west-2 --instance-id $INSTANCE_ID | jq ".AutoScalingInstances[].LifecycleState")

if grep -q "$LIFECYCLE" <<< "Warmed"; then
  exit 0
fi

## Bootstrap script below
```

But this would mean granting every node access to DescribeAutoscalingInstances which might be problematic

eytan-avisror:
Reached out to AWS, now the docs say:
```
[...]

PR #286, opened 2021-04-30
AL2 Support for AWS Warm Pools

eytan-avisror:
Adds to #277 

TBD:
- [x] Unit testing
- [x] Documentation


eytan-avisror:
```bash
33 scenarios (33 passed)
189 steps (189 passed)
31m12.218033102s
PASS
ok  	github.com/keikoproj/instance-manager/test-bdd	1872.780s [no tests to run]
```

eytan-avisror:
@backjo can you review when you get a chance?

api/v1alpha1/instancegroup_types.go
```diff
@@ -648,7 +670,7 @@ func (ig *InstanceGroup) Validate() error {
 			return err
 		}
 
-		if err := config.Validate(spec.Type); err != nil {
+		if err := config.Validate(spec.Type, spec.HasWarmPool()); err != nil {
```
backjo, line 673: Wonder if it makes more sense to add the validations for warm pool to spec validation (and also move the volume validation I added to spec validation, since it checks data from both spec and the config) - we don't seem to have a consistent approach to validation when we need to check data from spec and config - we have cases in both validation functions where we're checking both.

controllers/provisioners/eks/helpers.go
```diff
@@ -159,6 +159,15 @@ mount
 echo "{{ .Device}}    {{ .Mount }}    {{ .FileSystem | ToLower }}    defaults    0    2" >> /etc/fstab
 {{- end}}
 {{- end}}
+if [[ $(type -P $(which aws)) ]] && [[ $(type -P $(which jq)) ]] ; then
+	INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id)
```
backjo, line 163: Can we default to using IMDS v2? Some organizations might have a security policy around only allowing IMDS v2 to be used. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

api/v1alpha1/instancegroup_types.go
eytan-avisror: This would make sense, we can probably move warm pool validation to `(s *EKSSpec) Validate()` - we need access to both `EKSSpec` and `EKSConfiguration` so we can check if MixedInstancesPolicy or Spot is used.
Same thing for volumes.
I'll look into refactoring this

eytan-avisror:
Thanks for the review @backjo, I addressed the comments

backjo (approved):
Looks good - this is going to be really exciting to use!

PR #287, opened 2021-05-01
fix: correct merge logic for conditional rules

backjo:
this PR fixes a few bugs in the conditional rule logic for fields that can be merged, notably:
1. On 'merge' fields, the conditional value was being merged into the resource value before the default value was merged. This would cause the default value to take priority over the conditional value, which was not intended.
1. On 'merge' fields, the conditional value was not being set if a conflict was not present. 
1. On 'mergeOverride' fields, the conditional value was replacing the default value. Now, it is merged with the default value, with the conditional value taking priority. 

It also allows the userData field to be merged on the name key.

### DIFF
--- a/controllers/provisioners/config.go
+++ b/controllers/provisioners/config.go
@@ -34,12 +34,14 @@ var (
 	EKSTagsPath           = fmt.Sprintf("%v.tags", EKSConfigurationPath)
 	EKSVolumesPath        = fmt.Sprintf("%v.volumes", EKSConfigurationPath)
 	EKSLifecycleHooksPath = fmt.Sprintf("%v.lifecycleHooks", EKSConfigurationPath)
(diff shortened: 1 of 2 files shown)


eytan-avisror (approved):

controllers/provisioners/eks/helpers.go
```diff
@@ -286,6 +286,7 @@ func (ctx *EksInstanceGroupContext) GetAddedTags(asgName string) []*autoscaling.
 			tags = append(tags, ctx.AwsWorker.NewTag("k8s.io/cluster-autoscaler/node-template/label/kubernetes.io/os", "linux", asgName))
 		}
 
+
```
eytan-avisror, line 289: extra newline slipped through here

Issue #288, opened 2021-05-03
OsFamily missing validation in case wrong family provided


eytan-avisror:
In case where an instance-group is annotated with os-family of a wrong/unsupported value, we allow the operation and end up with empty user data:

```
scaling	detected drift	{"reason": "user-data has changed", "instancegroup": "ns/ig", "previousValue": "<old user-data>", "newValue": ""}
```

According to existing code, the desired behavior is to default to amazonlinux2, we should also check for allowed values:
https://github.com/keikoproj/instance-manager/blob/503d90c445366b4bc301420d85731c12332b1a16/controllers/provisioners/eks/eks.go#L133-L143

PR #289, opened 2021-05-03
fix: add validation for OsFamily annotation value

eytan-avisror:

Fixes: #288


vgunapati (approved):
LGTM

PR #290, opened 2021-05-04
feat: Support Windows for Warm Pools

backjo:
Adds to #277

### DIFF
--- a/controllers/provisioners/eks/helpers.go
+++ b/controllers/provisioners/eks/helpers.go
@@ -129,12 +129,12 @@ func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, k
   [string]$Lifecycle = Get-ASAutoScalingInstance $InstanceID | % { Echo $_.LifecycleState}
   if ($Lifecycle -like "*Warmed*") {
     Echo "Not starting Kubelet due to warmed state."
+    & C:\ProgramData\Amazon\EC2-Windows\Launch\Scripts\InitializeInstance.ps1 –Schedule
   } else {
     & $EKSBootstrapScriptFile -EKSClusterName {{ .ClusterName }} -KubeletExtraArgs '{{ .KubeletExtraArgs }}' 3>&1 4>&1 5>&1 6>&1
     {{range $post := .PostBootstrap}}{{$post}}{{end}}
   }
-</powershell>
-<persist>true</persist>`
+</powershell>`
 	case OsFamilyBottleRocket:
 		UserDataTemplate = `
 {{range $pre := .PreBootstrap}}{{$pre}}{{end}}
--- a/controllers/provisioners/eks/helpers.go
+++ b/controllers/provisioners/eks/helpers.go
@@ -124,9 +124,18 @@ func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, k
   [string]$EKSBinDir = "$env:ProgramFiles\Amazon\EKS"
   [string]$EKSBootstrapScriptName = 'Start-EKSBootstrap.ps1'
   [string]$EKSBootstrapScriptFile = "$EKSBinDir\$EKSBootstrapScriptName"
-  & $EKSBootstrapScriptFile -EKSClusterName {{ .ClusterName }} -KubeletExtraArgs '{{ .KubeletExtraArgs }}' 3>&1 4>&1 5>&1 6>&1
-  {{range $post := .PostBootstrap}}{{$post}}{{end}}
-</powershell>`
+  [string]$IMDSToken=(curl -UseBasicParsing -Method PUT "http://169.254.169.254/latest/api/token" -H @{ "X-aws-ec2-metadata-token-ttl-seconds" = "2
+1600"} | % { Echo $_.Content})
+  [string]$InstanceID=(curl -UseBasicParsing -Method GET "http://169.254.169.254/latest/meta-data/instance-id" -H @{ "X-aws-ec2-metadata-token" = "$IMDSToken"} | % { Echo $_.Content})
+  [string]$Lifecycle = Get-ASAutoScalingInstance $InstanceID | % { Echo $_.LifecycleState}
+  if ($Lifecycle -like "*Warmed*") {
+    Echo "Not starting Kubelet due to warmed state."
+  } else {
+    & $EKSBootstrapScriptFile -EKSClusterName {{ .ClusterName }} -KubeletExtraArgs '{{ .KubeletExtraArgs }}' 3>&1 4>&1 5>&1 6>&1
+    {{range $post := .PostBootstrap}}{{$post}}{{end}}
+  }
+</powershell>
+<persist>true</persist>`
 	case OsFamilyBottleRocket:
 		UserDataTemplate = `
 {{range $pre := .PreBootstrap}}{{$pre}}{{end}}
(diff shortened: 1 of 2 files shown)


controllers/provisioners/eks/helpers.go
```diff
@@ -124,8 +124,16 @@ func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, k
   [string]$EKSBinDir = "$env:ProgramFiles\Amazon\EKS"
   [string]$EKSBootstrapScriptName = 'Start-EKSBootstrap.ps1'
   [string]$EKSBootstrapScriptFile = "$EKSBinDir\$EKSBootstrapScriptName"
-  & $EKSBootstrapScriptFile -EKSClusterName {{ .ClusterName }} -KubeletExtraArgs '{{ .KubeletExtraArgs }}' 3>&1 4>&1 5>&1 6>&1
-  {{range $post := .PostBootstrap}}{{$post}}{{end}}
+  [string]$IMDSToken=(curl -UseBasicParsing -Method PUT "http://169.254.169.254/latest/api/token" -H @{ "X-aws-ec2-metadata-token-ttl-seconds" = "2
+1600"} | % { Echo $_.Content})
+  [string]$InstanceID=(curl -UseBasicParsing -Method GET "http://169.254.169.254/latest/meta-data/instance-id" -H @{ "X-aws-ec2-metadata-token" = "$IMDSToken"} | % { Echo $_.Content})
+  [string]$Lifecycle = Get-ASAutoScalingInstance $InstanceID | % { Echo $_.HealthStatus}
```
eytan-avisror, line 130: I think you need `LifecycleState` instead of `HealthStatus`

controllers/provisioners/eks/helpers.go
```diff
@@ -124,8 +124,16 @@ func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, k
...
+  if ($Lifecycle -like "*Warmed*") {
+    Echo "Not starting Kubelet due to warmed state."
```
eytan-avisror, line 132: You also need to delete the cloud-init state file here, otherwise user-data won't re-run on the next bootup

docs/EKS.md
```diff
@@ -414,7 +414,7 @@ You can configure your scaling group to use [AWS Warm Pools for Auto Scaling](ht
 Warm Pools is not officially supported with EKS, hence the following requirements exist in order to use it with EKS:
 - Your AMI must have `awscli` & `jq`, you can also install it in pre-bootstrap userdata.
```
eytan-avisror, line 415: We may want to modify this doc as well to say `awscli` & `jq` are needed in the case of AL2, pretty sure we also need awscli for windows nodes

controllers/provisioners/eks/helpers.go
backjo: Good catch! For Windows, this involved adding just a `<persist>true</persist>` block to user data.

docs/EKS.md
backjo: Clarified this - for Windows, the Get-ASAutoScalingInstance powershell cmdlet is the only dependency.

backjo:
Got around to testing this on the EKS recommended Windows AMI - it works!

Boot time for warm instances is around 2-3 minutes, compared to 7 for fresh instances. Might even be shorted without the custom PreBootstrap scripts that we have.

controllers/provisioners/eks/helpers.go
```diff
@@ -124,9 +124,17 @@ func (ctx *EksInstanceGroupContext) GetBasicUserData(clusterName, args string, k
...
+  [string]$InstanceID=(curl -UseBasicParsing -Method GET "http://169.254.169.254/latest/meta-data/instance-id" -H @{ "X-aws-ec2-metadata-token" = "$IMDSToken"} | % { Echo $_.Content})
+  [string]$Lifecycle = Get-ASAutoScalingInstance $InstanceID | % { Echo $_.LifecycleState}
+  if ($Lifecycle -like "*Warmed*") {
+    Echo "Not starting Kubelet due to warmed state."
+  } else {
+    & $EKSBootstrapScriptFile -EKSClusterName {{ .ClusterName }} -KubeletExtraArgs '{{ .KubeletExtraArgs }}' 3>&1 4>&1 5>&1 6>&1
+    {{range $post := .PostBootstrap}}{{$post}}{{end}}
+  }
+</powershell>
+<persist>true</persist>`
```
eytan-avisror, line 137: Can you only persist if the IF condition in line 130 is true? That way we only re-run if the boot was a warming.
This will be consistent with the behavior of AL2.
There might not be harm in always re-running user-data, but the default behavior is run-once

eytan-avisror:
> Got around to testing this on the EKS recommended Windows AMI - it works!
> 
> Boot time for warm instances is around 2-3 minutes, compared to 7 for fresh instances. Might even be shorted without the custom PreBootstrap scripts that we have.

Wow thats awesome, AL2 is kinda light weight so we only saw 1m instead of 2m approx, but still a 40-50% improvement.
Nice 💪

controllers/provisioners/eks/helpers.go
backjo: It looks like it - the docs say we can run `C:\ProgramData\Amazon\EC2-Windows\Launch\Scripts\InitializeInstance.ps1 –Schedule` to force an execution on the next startup.

backjo: Let me try it out real quick

backjo:
Yeah - I imagine BottleRocket will be even less of an improvement than AL2, since it's already starting up in 90 seconds or so.

eytan-avisror:
> Yeah - I imagine BottleRocket will be even less of an improvement than AL2, since it's already starting up in 90 seconds or so.

Well, AWS provisioning the instance takes around 45-60 seconds, booting up bottlerocket machine in 30 seconds, and making it ready in another 30 will be pretty damn impressive.

controllers/provisioners/eks/helpers.go
backjo: Works!

backjo:
I 

> > Yeah - I imagine BottleRocket will be even less of an improvement than AL2, since it's already starting up in 90 seconds or so.
> 
> Well, AWS provisioning the instance takes around 45-60 seconds (which you can potentially save by warming), booting up bottlerocket machine in 30 seconds, and making it ready in another 30 will be pretty damn impressive.

For real - I just clocked a BottleRocket node (without warm pool) from 
1) Manually updating the desired capacity on the ASG 
2) The Kubelet reporting to the API server 
3) The Node being ready on the API server

1-2 took 41 seconds
2-3 took 30 seconds

For a total of 71 seconds from ASG capacity change to being ready in K8S 😱

eytan-avisror:
> > > Yeah - I imagine BottleRocket will be even less of an improvement than AL2, since it's already starting up in 90 seconds or so.
> > 
> > 
> > Well, AWS provisioning the instance takes around 45-60 seconds (which you can potentially save by warming), booting up bottlerocket machine in 30 seconds, and making it ready in another 30 will be pretty damn impressive.
> 
> For real - I just clocked a BottleRocket node (without warm pool) from
> 
>     1. Manually updating the desired capacity on the ASG
> 
>     2. The Kubelet reporting to the API server
> 
>     3. The Node being ready on the API server
> 
> 
> 1-2 took 41 seconds
> 2-3 took 30 seconds
> 
> For a total of 71 seconds from ASG capacity change to being ready in K8S 😱 .
> 
> Given the design of BottleRocket with it's read-only root file system and the already bleeding-fast startup times, it probably doesn't make sense to even attempt to support warm pools for it.

Might reduce that by a good 20-30 seconds with warm pools :)
Still worth exploring given the cost of running this is almost free.

BTW, did you push the change for `<persist>`?

backjo:
Just pushed the persist change

eytan-avisror (approved):
🏆

PR #303, opened 2021-05-11
fix: Ignore mixedInstancePolicy if type is LaunchConfig

### LINKED ISSUE #301: Validation: block using mixedInstancePolicy with LaunchConfiguration as default
When a CR does not mention spec.Type it defaults to LaunchConfiguration, however if the CR mentions mixedInstancePolicy it would keep trying to reconcile indefinitely.

We should either ignore the MixedInstancePolicySpec in this case, or error out the reconcile with a proper log message.

eytan-avisror:

Fixes #301 

This ignores mixedInstancePolicy spec if it is used along with LaunchConfiguration as spec type.

### DIFF
--- a/api/v1alpha1/instancegroup_types.go
+++ b/api/v1alpha1/instancegroup_types.go
@@ -537,7 +537,7 @@ func (c *EKSConfiguration) Validate() error {
 	for _, v := range c.Volumes {
 
 		if v.Iops != 0 && !common.ContainsEqualFold(awsprovider.AllowedVolumeTypesWithProvisionedIOPS, v.Type) {
-			return errors.Errorf("cannot apply IOPS configuration for volumeType, only types ['io1','io2','gp3'] supported", "volumeType", v.Type)
+			return errors.Errorf("cannot apply IOPS configuration for volumeType '%v', only types '%v' supported", v.Type, awsprovider.AllowedVolumeTypesWithProvisionedIOPS)
 		}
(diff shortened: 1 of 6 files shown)


api/v1alpha1/instancegroup_types.go
```diff
@@ -392,6 +392,13 @@ func (s *EKSSpec) Validate() error {
 		s.Type = LaunchConfiguration
 	}
 
+	if s.Type == LaunchConfiguration {
+		if s.EKSConfiguration.MixedInstancesPolicy != nil {
+			log.Info("cannot use mixedInstancesPolicy with LaunchConfiguration, will ignore provided mixedInstancePolicy")
```
backjo, line 397: Not sure if this log is going to be very useful - does it contain any identifying information for the IG?

Might also make sense to event this on the IG, for users that aren't cluster operators.

eytan-avisror: Yeah, we should pass the IG name from the top level `Validate()` so we can add the IG name.
Regarding events - yeah, that's a great idea to have events for any IG error associated with the IG for visibility - we might want to do that as a separate effort though across the board.

eytan-avisror: Created https://github.com/keikoproj/instance-manager/issues/304 to address events

backjo (approved):

PR #305, opened 2021-05-15
feat: add support for instance metadata options

backjo:
Closes #305


controllers/provisioners/eks/scaling/launchconfig.go
```diff
@@ -282,6 +293,17 @@ func (lc *LaunchConfiguration) blockDeviceList(volumes []v1alpha1.NodeVolume) []
 	return sortConfigDevices(devices)
 }
 
+func (lc *LaunchConfiguration) metadataOptions(input *v1alpha1.MetadataOptions) *autoscaling.InstanceMetadataOptions {
```
eytan-avisror, line 296: Current pattern is to place these kind of constructors in awsprovider, could we do the same in this case? There is no real need for this to be a method of LaunchConfiguration -  would be good if we can merge #296 first to avoid the conflict

controllers/provisioners/eks/scaling/launchtemplate.go
```diff
@@ -345,6 +356,28 @@ func (lt *LaunchTemplate) launchTemplatePlacementRequest(input *v1alpha1.Placeme
 	return lt.LaunchTemplatePlacementRequest(input.AvailabilityZone, input.HostResourceGroupArn, input.Tenancy)
 }
 
+func (lt *LaunchTemplate) metadataOptions(input *v1alpha1.MetadataOptions) *ec2.LaunchTemplateInstanceMetadataOptions {
```
eytan-avisror, line 359: Same for these two methods

eytan-avisror:
LG, added suggestions to move some code to awsprovider

eytan-avisror:
@backjo nevermind, I see actually in the cause of launchtemplate we are placing these constructors within the package.
We can refactor later if it makes sense, for now it's aligned with current patterns.

eytan-avisror (approved):

PR #309, opened 2021-05-19
Pass API Endpoint & CA Cert directly to bootstrap.sh

### LINKED ISSUE #306: API Retries on bootstrap.sh
The bootstrap.sh script has a flag `--aws-api-retry-attempts` that defaults to 3.
If the account is getting throttled heavily, this may cause nodes to fail joining the cluster.
We should set a higher number e.g. 12 to avoid this scenario as much as possible

eytan-avisror:
Fixes #306 

This PR uses the payload from DescribeCluster to feed CA & API endpoint directly to userdata in order to avoid making those calls and potentially getting throttled on user-data.

This also fixes the argument passing in the case of Windows so that we wrap the Kubelet arguments same as we do for Linux

### TBD
- [x] Test Linux AMI
- [x] Test Windows AMI

### DIFF
--- a/controllers/provisioners/eks/helpers.go
+++ b/controllers/provisioners/eks/helpers.go
@@ -515,9 +515,10 @@ func (ctx *EksInstanceGroupContext) GetBootstrapArgs() string {
 		bootstrapOptions = ctx.GetComputedBootstrapOptions()
 		state            = ctx.GetDiscoveredState()
 		osFamily         = ctx.GetOsFamily()
+		cluster          = state.GetCluster()
+		clusterIP        = ctx.AwsWorker.GetDNSClusterIP(cluster)
 	)
 	var sb strings.Builder
-
 	switch strings.ToLower(osFamily) {
 	case OsFamilyWindows:
 		if state.Cluster != nil {
@@ -532,6 +533,9 @@ func (ctx *EksInstanceGroupContext) GetBootstrapArgs() string {
 		if state.Cluster != nil {
 			sb.WriteString(fmt.Sprintf("--b64-cluster-ca %v ", aws.StringValue(state.Cluster.CertificateAuthority.Data)))
 			sb.WriteString(fmt.Sprintf("--apiserver-endpoint %v ", aws.StringValue(state.Cluster.Endpoint)))
+			if !common.StringEmpty(clusterIP) {
+				sb.WriteString(fmt.Sprintf("--dns-cluster-ip %v ", clusterIP))
+			}
 		}
 
 		sb.WriteString(fmt.Sprintf("--kubelet-extra-args '%v'", ctx.GetKubeletExtraArgs()))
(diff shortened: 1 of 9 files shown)


controllers/provisioners/eks/helpers.go
```diff
@@ -513,13 +513,30 @@ func (ctx *EksInstanceGroupContext) GetComputedBootstrapOptions() *v1alpha1.Boot
...
+			sb.WriteString(fmt.Sprintf("-Base64ClusterCA %v ", aws.StringValue(state.Cluster.CertificateAuthority.Data)))
+			sb.WriteString(fmt.Sprintf("-APIServerEndpoint %v ", aws.StringValue(state.Cluster.Endpoint)))
+		}
+		sb.WriteString(fmt.Sprintf("-KubeletExtraArgs '%v'", ctx.GetKubeletExtraArgs()))
+	case OsFamilyAmazonLinux2:
+		if bootstrapOptions != nil && bootstrapOptions.MaxPods > 0 {
+			sb.WriteString("--use-max-pods false ")
+		}
+		if state.Cluster != nil {
+			sb.WriteString(fmt.Sprintf("--b64-cluster-ca %v ", aws.StringValue(state.Cluster.CertificateAuthority.Data)))
```
backjo, line 533: There's some tricky behavior in the EKS AL2 AMI that we might want to solve for before merging this change.
See https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L332

Basically, if the cluster ca and api server endpoint are defined, the AMI bypasses calling EKS - which is what we want! Unfortunately, that call also grabs the serviceIpv4Cidr field, which is used later in the bootstrap script to calculate the default ip that should be used for DNS (the kube-dns/core-dns service ip basically). We probably want to replicate this logic and inject the dns ip address in case the user is relying on this behavior.

eytan-avisror: Good catch, will look into this

eytan-avisror: @backjo take a look again, I added discovery of cluster ip in the case of AL2

backjo: The changes look pretty good! If I might suggest one more change - looking up the IPv4 CIDR range of the node that IM is running on adds a somewhat implicit assumption that Instance Manager is running in the same VPC - which I think is generally a fair assumption, though it could make debugging with a remote cluster from an IDE a bit more difficult :P 

I think a cleaner approach to determine the VPC CIDR range for the nodes might be to lookup the VPC attached to the EKS cluster - and use the CIDR range associated with it for the fallback logic.

eytan-avisror: Yeah.. technically we can remove the fallback all together, since reconcile depends on getting cluster information.
I think it's safe to assume we will always have the ServiceCIDR and remove the metadata fallback logic. WDYT?

backjo: I'm fine with it - the only open question I have is whether ServiceCIDR is always defined - I know that it was only added in October 2020 - and all of our clusters are newer than that - so could use some help verifying that even for older clusters, the value is defined.

eytan-avisror: I can confirm I have clusters from 2019 which have this populated.
Makes sense when they add a new API to make sure it's populated across all objects otherwise this would be impossible to use

eytan-avisror: I've also removed the ec2metadata discovery fallback since we guarantee to always have the cluster payload or fail the reconcile.

backjo: Cool - sounds good!

backjo (approved):

PR #315, opened 2021-06-09
fix: add addl tags for scale to 0 support for windows nodes

backjo:


controllers/provisioners/eks/helpers.go
```diff
@@ -302,6 +302,10 @@ func (ctx *EksInstanceGroupContext) GetAddedTags(asgName string) []*autoscaling.
 		switch strings.ToLower(osFamily) {
 		case OsFamilyWindows:
 			tags = append(tags, ctx.AwsWorker.NewTag("k8s.io/cluster-autoscaler/node-template/label/kubernetes.io/os", "windows", asgName))
+			tags = append(tags, ctx.AwsWorker.NewTag("k8s.io/cluster-autoscaler/node-template/label/kubernetes.io/arch", "amd64", asgName))
+			instanceTypeNetworkInfo := awsprovider.GetInstanceTypeNetworkInfo(ctx.GetDiscoveredState().GetInstanceTypeInfo(), configuration.InstanceType)
```
eytan-avisror, line 306: We also call this in `GetComputedBootstrapOptions`, should we simply make sure this call is made in CloudDiscovery, and then stored in DiscoveredState?

backjo: yep - good call

eytan-avisror: Oh NVM, I thought this was an API call, looks like it's just a helper function 👍 
In this case can we just move `ctx.GetDiscoveredState().GetInstanceTypeInfo()` to top var declaration for order/OCD purposes 😀

controllers/provisioners/eks/helpers.go
```diff
@@ -302,6 +302,10 @@ func (ctx *EksInstanceGroupContext) GetAddedTags(asgName string) []*autoscaling.
...
+			numberOfIps := *instanceTypeNetworkInfo.Ipv4AddressesPerInterface - 1
```
eytan-avisror, line 307: Should we use this instead:
```golang
numberOfIps := aws.Int64Value(instanceTypeNetworkInfo.Ipv4AddressesPerInterface) - 1
```

eytan-avisror: It's also possible for `GetInstanceTypeNetworkInfo` to return nil, so we might want a nil check to avoid a nil pointer? Not sure if feasible for this to happen

eytan-avisror (approved):
💯

PR #319, opened 2021-06-29
fix: calculate cluster dns correctly for non /16 blocks

### LINKED ISSUE #318: Cluster DNS calculation assumes that service cidr range is /16
BUG REPORT 
**What happened**:
Pods running on AL2 nodes started having DNS issues when running the latest version of instance manager. Upon investigation, 
this appears to be because the EKS Cluster CIDR block is a /12 range - which causes this string replacement to fail https://github.com/keikoproj/instance-manager/blob/master/controllers/providers/aws/eks.go#L222. 

Pods should have working cluster DNS.
Have an EKS cluster with a ServiceCidr range that is not a /16 and run AL2 nodes.

**Environment**:

- Kubernetes version:

```bash
kubectl version -o yaml
```

- InstanceGroup status:

```bash
```

- controller logs:

```bash
```

backjo:
Closes #318 


eytan-avisror (approved):
LG

PR #324, opened 2021-07-22
release 0.12.2

backjo:


eytan-avisror:
@backjo probably more appropriate to release as 0.13.0 as it contains features and not just fixes.
Also, for some reason last nightly bdd failed, can we wait for the bdd to pass? alternatively we can manually run it

backjo:
Good point - let me update that. And yeah, no rush.

Issue #321, opened 2021-07-15
Upgrade approval for changes to Instance Groups


preflightsiren:
**Feature request**

When an `InstanceGroup` is modified, there should be an approval process that controls when the `InstanceGroup` change is triggered.

See Certificate-Manager for an example: https://cert-manager.io/docs/concepts/certificaterequest/#approval

strawman workflow:

1. Change to `InstanceGroup`
2. `UpgradeRequest` CR is created
3. Upgrade operator marks `UpgradeRequest` `Approved`
4. Instance Manager creates the Custom Resource (eg. Argo workflow or RollingUpgrade)
5. Modification continues as currently exists
6.

preflightsiren:
inspired by #320

PR #327, opened 2021-08-05
WIP: 320 automated ami upgrades + upgrade locking

preflightsiren:


preflightsiren:
scenarios tested:
* Adding locked annotation sets state to `Locked`
* Removing locked annotation set state to `Ready`
* Modifying IG in `Locked` state does not enter ReconcileModifying
* Deleting IG in `Locked` state deletes AWS resources, but not the `IG` - whoops.
* Creating IG in `Locked` state creates AWS resources; IG does not transition through states. - whoops.

api/v1alpha1/instancegroup_types.go
```diff
@@ -47,8 +47,9 @@ const (
 	ReconcileModified  ReconcileState = "ReconcileModified"
 
 	// End States
-	ReconcileReady ReconcileState = "Ready"
-	ReconcileErr   ReconcileState = "Error"
+	ReconcileLocked ReconcileState = "Locked"
```
eytan-avisror, line 50: Not sure "Locked" should be a reconcile state - the intention is to lock upgrades, not reconciles.

eytan-avisror:
Thanks for the PR @preflightsiren !
I think we need to consider the behavior/design we want here with locking - whether we lock all reconciles or just don't enter the upgrade path.

I think it would have less implications if we only avoid upgrades (e.g. a condition check before triggering upgrades) vs. a reconcile state.

Otherwise we need to consider all other states and transitions between them from Locked state.

preflightsiren:
Hey mate, thanks for taking a look. From observation each change to the IG sets the state back to `Init` meaning we only need to worry about the transitions into `Locked`, rather than worrying about how state transitions in and out of those final states? (I think we may be agreeing on this point)

I would want a `Locked` reconcile state to show operators why a change isn't being applied. Rather it show `Ready` with a log message or similar. 

In its current form it only locks updates and upgrades - although currently implemented it's not a great experience with creates/deletes. 

Next steps are to address the create/delete flows for locked IGs, and implement the imageID lookup.

api/v1alpha1/instancegroup_types.go
preflightsiren: I'm missing something here, what's an example of a reconcile that doesn't result in an update/upgrade?

from my understanding; each change to the `CR` will trigger the state to go to `Init` eg.

```
2021-08-05T22:22:09.340+1000    INFO    v1alpha1        state transition occured        {"instancegroup": "instance-manager/platform-us-west-2b", "state": "Init", "previousState": "Locked"}
```

and then flow following:

```
func HandleReconcileRequest(d CloudDeployer) error 
```

I've used the reconcile state to mark that:
1. The `IG` should show it's `Locked`
2. The AWS resources should not be modified by preventing entering the `ReconcileInitUpdate`, `ReconcileInitUpgrade` 

remaining questions:
1. Is there a difference between a reconcile and an upgrade/update?
2. Is my understanding of the reconcile state flow correct?
3. where does it miss the design we've discussed so far?

thanks for your help :)

controllers/providers/aws/ssm.go
```diff
@@ -0,0 +1,56 @@
...
+		log.V(1).Info("AWS API call",
+			"cacheHit", cache.IsCacheHit(ctx),
+			"service", r.ClientInfo.ServiceName,
+			"operation", r.Operation.Name,
+		)
+	})
+	return ssm.New(sess)
+}
+
+func (w *AwsWorker) GetEksLatestAmi(clusterName string) (string, error) {
```
backjo, line 39: This function will need to handle variability across OS families and architectures - i.e. it should be able to get the latest AMIs for:

- Bottlerocket OS running x86_64
- Bottlerocket OS running ARM
- Amazon Linux 2 running x86_64
- Amazon Linux 2 running ARM
- Windows 2019 running x86_64

Larger organizations will also generally prefer the ability to customize this - i.e. have a custom SSM path to point to.

api/v1alpha1/instancegroup_types.go
eytan-avisror: For showing something on kubectl we can also embed `locked: true/false` in status and display it as a new column if we wanted - we don't have to use reconcile state, but I think from display perspective it makes sense.

Your questions:
1. Reconcile-update is for non node rotating changes, e.g. add an EC2 tag. or basically any change that is not on the launch template / launch config. reconcile-upgrade would be changes that cause a drift in the launch config / template.
2. Yes, I think so, the interface shows the flow pretty clearly - if you put `Lock` at first it means you are blocking other flows from executing.
3. I think the original intention was to only block node-rotation. However with this design we are blocking updates to ASG which do not cause node rotations - while that also might be a nice feature we should consider the use cases around it, and also reconsider the annotation name to simply be `"instancemgr.keikoproj.io/locked"` in this case.

controllers/providers/aws/ssm.go
preflightsiren: All good points. Do you think they should be implemented in a v1 of this feature?
Or can we implement what's needed today and expand it to fit need?

backjo: I'd prefer to implement as V1 - or an _extremely_ fast follow - at least for detection based off of os-family and architecture. I don't think custom paths is required for v1.

preflightsiren: Correct me if I'm wrong, but doesn't that information come _from_ the AMI? So we'd need to get the user to pass in somehow what architecture they want to use?

Did you have a suggestion on how that should work?

The simplest and ugliest way I can think of is to encode that information into the string; have instead of `latest`, `latest-al2-x86`, `latest-bottlerocket-arm64` or `latest-windows-x86`

backjo: Not necessarily - the IG will be annotated with the os-family being used (defaults to al2). For architecture, it can be derived from the type of instance that is being used - i.e. M5 family will always be x86, M6g family will always be arm

preflightsiren: just found `instancemgr.keikoproj.io/os-family` ok, that should be easy - assuming amazon provide a latest ami for those variations :)

controllers/providers/aws/ssm.go
```diff
@@ -0,0 +1,56 @@
...
+	log.Info("GetEksLatestAmi")
+	version, err := w.GetEksClusterVersion(clusterName)
```
eytan-avisror, line 41: We should already have the cluster version since we make the call separately in CloudDiscovery.
Can you decouple it? e.g. take version as a parameter.

preflightsiren: I'll double check - I created this because I believe we call `DescribeCluster` for a different purpose

eytan-avisror: My point is that we already make the DescribeCluster call on reconcile every time and it would be counter productive to make that call again (although we have caching) - we call it to get the cluster version actually.

So you can pass in the cluster from cached state into `GetEksLatestAmi` and avoid the extra API call.

https://github.com/keikoproj/instance-manager/blob/fe8b56f407f75ca71dbdb8e08892b6fece610e0a/controllers/provisioners/eks/cloud.go#L165-L169

You can use the following method on DiscoveredState:
https://github.com/keikoproj/instance-manager/blob/fe8b56f407f75ca71dbdb8e08892b6fece610e0a/controllers/provisioners/eks/cloud.go#L300-L305

api/v1alpha1/instancegroup_types.go
preflightsiren: Whats the flow you would expect for a locked IG?

Should we bypass the upgrade flow?
`Init` -> `InitUpgrade` -> `Ready` 

by adding a condition on the `UpgradeNodes()` call: 

```
if d.GetState() == v1alpha.ReconcileInitUpgrade && !d.Locked() {
  err = d.UpgradeNodes()
```

Should we

`Init` -> `InitUpgrade` -> `ReconcileModifying` -> `ReconcileModified` -> `Ready` 

and shortcut the flow in `UpgradeNodes()`:

```
func (ctx *EksInstanceGroupContext) UpgradeNodes() error {
  var (
... snip ...
  )

  if ctx.Configuration.InstanceGroup.Locked() {
    return nil
  }
```

or something else?

eytan-avisror: What if we just short circuit from the interface? would that work?
e.g. 
```golang
if d.Locked() {
  d.SetState(v1alpha.ReconcileLocked)
  return nil
}
```

controllers/providers/aws/ssm.go
preflightsiren: I have to choose between Windows Cofe and Windows Full; any preferences?

backjo: Core

preflightsiren:
Just looking through https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-public-parameters-eks.html

The ssm GetParameter response contains `LastModifiedDate`. This might be useful to create policies to allow teams to set different delays updating. Eg. Dev clusters immediately get new AMIs, staging clusters after 48h, prod after 1week. 

What do you think?

preflightsiren:
success with multiple OS's

```
NAME                                          STATUS   ROLES                           AGE     VERSION
ip-172-29-72-68.us-west-2.compute.internal    Ready    <none>                          14d     v1.19.6-eks-49a6c0
ip-172-29-73-206.us-west-2.compute.internal   Ready    platform-us-west-2a             22h     v1.19.6-eks-49a6c0
ip-172-29-73-85.us-west-2.compute.internal    Ready    platform-us-west-2a             22h     v1.19.6-eks-49a6c0
ip-172-29-75-149.us-west-2.compute.internal   Ready    amazonlinux2-arm64-us-west-2b   5h57m   v1.19.13-eks-8df270
ip-172-29-75-175.us-west-2.compute.internal   Ready    bottlerocket-us-west-2b         5h57m   v1.19.13
ip-172-29-75-196.us-west-2.compute.internal   Ready    windows-us-west-2b              5h52m   v1.19.8-eks-fd6eea
ip-172-29-75-238.us-west-2.compute.internal   Ready    amazonlinux2-us-west-2b         5h57m   v1.19.13-eks-8df270
ip-172-29-77-46.us-west-2.compute.internal    Ready    <none>                          14d     v1.19.6-eks-49a6c0
```

preflightsiren:
Updating a locked instance group:
```
2021-08-19T21:23:06.496+1000    INFO    v1alpha1        state transition occured        {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "state": "InitUpdate", "previousState": "Init"}
2021-08-19T21:23:06.496+1000    INFO    v1alpha1        state transition occured        {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "state": "ReconcileModifying", "previousState": "InitUpdate"}
2021-08-19T21:23:06.496+1000    DEBUG   aws-provider    AWS API call    {"cacheHit": true, "service": "ssm", "operation": "GetParameter"}
2021-08-19T21:23:06.496+1000    INFO    aws-provider    Latest EKS AMI ID:      {"value": "ami-0a8ad9f45d082a280"}
2021-08-19T21:23:06.497+1000    INFO    scaling drift not detected      {"instancegroup": "instance-manager/amazonlinux2-us-west-2b"}
2021-08-19T21:23:06.497+1000    INFO    controllers.instancegroup.eks   node rotation required  {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "scalingconfig": "uw2d-akp-scole-instance-manager-amazonlinux2-us-west-2b-20210819050930"}
2021-08-19T21:23:06.871+1000    DEBUG   aws-provider    AWS API call    {"cacheHit": false, "service": "autoscaling", "operation": "CreateOrUpdateTags"}
2021-08-19T21:23:06.871+1000    INFO    controllers.instancegroup.eks   updated scaling group tags      {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "scalinggroup": "uw2d-akp-scole-instance-manager-amazonlinux2-us-west-2b"}
2021-08-19T21:23:06.871+1000    INFO    controllers.instancegroup.eks   bootstrapping arn to aws-auth   {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "arn": "arn:aws:iam::026026080755:role/akp-eks-node-us-west-2"}
2021-08-19T21:23:07.229+1000    INFO    controllers.instancegroup.eks   waiting for node readiness conditions   {"instancegroup": "instance-manager/amazonlinux2-us-west-2b"}
2021-08-19T21:23:07.229+1000    INFO    controllers.instancegroup.eks   desired nodes are ready {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "instances": "i-0dd10ab2bbbfcbacc"}
2021-08-19T21:23:07.229+1000    INFO    v1alpha1        state transition occured        {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "state": "ReconcileModified", "previousState": "ReconcileModifying"}
2021-08-19T21:23:07.229+1000    INFO    v1alpha1        state transition occured        {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "state": "InitUpgrade", "previousState": "ReconcileModified"}
2021-08-19T21:23:07.229+1000    INFO    v1alpha1        state transition occured        {"instancegroup": "instance-manager/amazonlinux2-us-west-2b", "state": "Locked", "previousState": "InitUpgrade"}
```

eytan-avisror:
@preflightsiren thanks, I am OOO until mid next week - will review as soon as I can. Suggest in the meanwhile you consider adding a functional test? Or at least make sure they are passing considering this is a fairly large change.

controllers/providers/aws/aws.go
```diff
@@ -238,3 +241,12 @@ func GetInstanceTypeNetworkInfo(instanceTypes []*ec2.InstanceTypeInfo, instanceT
 	}
 	return nil
 }
+
+func GetInstanceTypeArchetectures(instanceTypes []*ec2.InstanceTypeInfo, instanceType string) []*string {
```
backjo, line 245: Archetectures -> Architectures

controllers/providers/aws/aws.go
```diff
@@ -238,3 +241,12 @@ func GetInstanceTypeNetworkInfo(instanceTypes []*ec2.InstanceTypeInfo, instanceT
...
+	for _, instanceTypeInfo := range instanceTypes {
```
backjo, line 246: probably makes sense to move the instanceType lookup into a helper function since we're doing the same thing in the function above.

controllers/providers/aws/ssm.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+var (
+	EksAmis = map[string]architectureMap{
+		"amazonlinux2": architectureMap{
+			"x86_64": EksOptimisedAmiPath,
+			"arm64":  EksOptimisedAmazonLinux2ARM64,
+		},
+		"bottlerocket": architectureMap{
+			"x86_64": EksOptimisedBottlerocket,
```
backjo, line 32: BottleRocket should have an arm64 ami as well

controllers/provisioners/eks/create.go
```diff
@@ -62,6 +62,14 @@ func (ctx *EksInstanceGroupContext) Create() error {
 		configName = fmt.Sprintf("%v-%v", ctx.ResourcePrefix, common.GetTimeString())
 	}
 
+	if configuration.Image == "latest" {
```
backjo, line 65: can we make latest a constant?

controllers/provisioners/eks/helpers.go
```diff
@@ -1173,3 +1173,40 @@ func (ctx *EksInstanceGroupContext) GetDesiredMixedInstancesPolicy(name string)
...
+func (ctx *EksInstanceGroupContext) GetEksLatestAmi() (string, error) {
+	var (
+		instanceGroup = ctx.GetInstanceGroup()
+		state         = ctx.GetDiscoveredState()
+		configuration = instanceGroup.GetEKSConfiguration()
+	)
+	clusterVersion := state.GetClusterVersion()
+	annotations := instanceGroup.GetAnnotations()
+	var OSFamily string
+	if annotations[OsFamilyAnnotation] != "" {
```
backjo, line 1200: could probably move this logic into a helper function - I believe it exists elsewhere as well

backjo (commented):
Overall, looks pretty good! A few minor comments - and a +1 to Eytan's suggestion for a functional test. 

Could also use some documentation of the lock down annotation.

api/v1alpha1/instancegroup_types.go
```diff
@@ -87,6 +88,8 @@ const (
 
 	DockerRuntime     ContainerRuntime = "dockerd"
 	ContainerDRuntime ContainerRuntime = "containerd"
+
+	UpgradeLockedAnnotationKey = "instancemgr.keikoproj.io/lock-upgrades"
```
eytan-avisror, line 92: Can we move this to `eks.go` with other annotations?
https://github.com/keikoproj/instance-manager/blob/master/controllers/provisioners/eks/eks.go#L36-L43

api/v1alpha1/instancegroup_types.go
```diff
@@ -392,6 +395,10 @@ func (ig *InstanceGroup) GetUpgradeStrategy() *AwsUpgradeStrategy {
 func (ig *InstanceGroup) SetUpgradeStrategy(strategy AwsUpgradeStrategy) {
 	ig.Spec.AwsUpgradeStrategy = strategy
 }
+func (ig *InstanceGroup) Locked() bool {
+	annotations := ig.GetAnnotations()
+	return annotations[UpgradeLockedAnnotationKey] == "true"
```
eytan-avisror, line 400: You can probably use `HasAnnotation` from https://github.com/keikoproj/instance-manager/blob/c0eb25d65856ca1261f70465c56950f4bfc0cdc3/controllers/providers/kubernetes/utils.go#L119

It has a bit more safety for case, etc.

controllers/providers/aws/ssm.go
```diff
@@ -0,0 +1,71 @@
...
+func (w *AwsWorker) GetEksLatestAmi(OSFamily string, arch string, kubernetesVersion string) (string, error) {
+	input := &ssm.GetParameterInput{
+		Name: aws.String(fmt.Sprintf(EksAmis[OSFamily][arch], kubernetesVersion)),
+	}
+
+	output, err := w.SsmClient.GetParameter(input)
+	if err != nil {
+		return "", err
+	}
+	log.Info("Latest EKS AMI ID:", "value", output.Parameter.Value)
```
eytan-avisror, line 69: We might want to move logging up to caller since this is more of a helper function

controllers/provisioners/eks/create.go
eytan-avisror: Also good if you can use `strings.EqualFold` to avoid case sensitivity issues

controllers/provisioners/eks/helpers.go
eytan-avisror: See `HasAnnotation` helper

controllers/providers/aws/aws.go
```diff
@@ -238,3 +241,12 @@ func GetInstanceTypeNetworkInfo(instanceTypes []*ec2.InstanceTypeInfo, instanceT
...
+		if aws.StringValue(instanceTypeInfo.InstanceType) == instanceType {
+			return instanceTypeInfo.ProcessorInfo.SupportedArchitectures
```
eytan-avisror, line 248: Would be good if this can return a string value slice. you can use `return aws.StringValueSlice(instanceTypeInfo.ProcessorInfo.SupportedArchitectures)`

controllers/provisioners/eks/helpers.go
```diff
@@ -1173,3 +1173,40 @@ func (ctx *EksInstanceGroupContext) GetDesiredMixedInstancesPolicy(name string)
 
 	return policy
 }
+
+func filterSupportedArch(architectures []*string) string {
```
eytan-avisror, line 1177: Once you return StringValueSlice in the previous call to get architecture, you can refactor this function a bit.
e.g., you can simply use ContainsEqualFold from here:
https://github.com/keikoproj/instance-manager/blob/c0eb25d65856ca1261f70465c56950f4bfc0cdc3/controllers/common/utils.go#L80

You can also make `SupportedArchitectures` available at a package level in case it needs to be used elsewhere

controllers/provisioners/eks/update.go
```diff
@@ -64,6 +64,14 @@ func (ctx *EksInstanceGroupContext) Update() error {
 	}
 	instanceProfile := state.GetInstanceProfile()
 
+	if configuration.Image == "latest" {
```
eytan-avisror, line 67: Seems this code is repetative in `update.go` and `create.go`, should we simply do this check in `cloud.go` and overwrite the field internally with the resolved value?

docs/examples/EKS.md
```diff
@@ -177,3 +178,25 @@ spec:
 ```
 
 if you do not provide these fields, a role will be created for your instance-group by the controller (will require IAM access).
+
+
+### Auto-Updating AWS EKS Optimised AMI ID
```
eytan-avisror, line 183: We may want to add a comment that there might be some delay due to reconcile not being triggered.
E.g. new AMI is out, TTL for GetParameter is expired so we will get the new AMI, but nothing is triggering the reconcile. 
Until the new reconcile is triggered upgrade will not happen.

This is good if you use the locking mechanism, because removing the lock annotation will trigger the reconcile, but otherwise it might take a while until that happens.

In addition we need to add the newly added AWS SSM permissions to documentation here:
https://github.com/keikoproj/instance-manager/blob/master/docs/INSTALL.md

preflightsiren:
Good feedback; haven't had a chance yet, but I'll add functional tests, and I'll look at providing a ssm parameter path override annotation as well

api/v1alpha1/instancegroup_types.go
preflightsiren: How important is this move? a direct approach results in a circular dependency between v1alpha1 and eks - could try and rework it, but it seems non-trivial.

api/v1alpha1/instancegroup_types.go
preflightsiren: Same as above, v1alpha is included by `kubernetes/utils.go` so creates a circular dependency.

preflightsiren: I've duplicated the functionality into the `Locked()` method.

controllers/providers/aws/ssm.go
preflightsiren: Following https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id-bottlerocket.html Amazon only document a x86_64 AMI; I've tried to follow the expected path and got a result - fingers crossed continues to be supported :)

```
{
    "Parameter": {
        "Name": "/aws/service/bottlerocket/aws-k8s-1.19/arm64/latest/image_id",
        "Type": "String",
        "Value": "ami-09bfe0d7ac581d319",
        "Version": 10,
        "LastModifiedDate": "2021-08-07T03:26:47.119000+10:00",
        "ARN": "arn:aws:ssm:us-west-2::parameter/aws/service/bottlerocket/aws-k8s-1.19/arm64/latest/image_id",
        "DataType": "text"
    }
}
```

docs/examples/EKS.md
preflightsiren: > We may want to add a comment that there might be some delay due to reconcile not being triggered.
> E.g. new AMI is out, TTL for GetParameter is expired so we will get the new AMI, but nothing is triggering the reconcile.

I thought kubernetes objects get periodically reconciled?

eytan-avisror: They do, but I think the default kubebuilder period is 24hrs, so that would be significant delay.
There are other thing that can trigger reconcile such as spot recommendations etc.

api/v1alpha1/instancegroup_types.go
eytan-avisror: Yeah, we can leave there in that case, it was just a nit

api/v1alpha1/instancegroup_types.go
eytan-avisror: 👍

docs/examples/EKS.md
preflightsiren: from: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/config/v1alpha1#ControllerConfigurationSpec
```
// SyncPeriod determines the minimum frequency at which watched resources are
	// reconciled. A lower period will correct entropy more quickly, but reduce
	// responsiveness to change if there are many watched resources. Change this
	// value only if you know what you are doing. Defaults to 10 hours if unset.
	// there will a 10 percent jitter between the SyncPeriod of all controllers
	// so that all controllers will not send list requests simultaneously.
	// +optional
	SyncPeriod *metav1.Duration `json:"syncPeriod,omitempty"`
```

eytan-avisror: That's it - we don't set it, so have the 10hr default.
Wonder if we should make this a global controller flag (we can do that later if we want), but would be good to add a small comment in the docs on this if this is the expected behavior (may take up to 10hrs + API TTL to get upgrade trigger)

preflightsiren:
woo, got the BDD tests running, all passed:

```
33 scenarios (33 passed)
189 steps (189 passed)
38m38.269353018s
PASS
ok      github.com/keikoproj/instance-manager/test-bdd  2318.760s [no tests to run]
```

preflightsiren:
Added BDD tests for locking the group

preflightsiren:
```
37 scenarios (37 passed)
214 steps (214 passed)
48m3.043404152s
PASS
ok      github.com/keikoproj/instance-manager/test-bdd  2883.513s [no tests to run]
```

now includes tests for locking and latest.

controllers/providers/kubernetes/utils.go
```diff
@@ -116,7 +116,14 @@ func AddAnnotation(u *unstructured.Unstructured, key, value string) {
...
 
-func HasAnnotation(annotations map[string]string, key, value string) bool {
+func HasAnnotation(annotations map[string]string, key string) bool {
+	if _, ok := annotations[key]; ok {
+		return true
+	}
+	return false
+}
+
+func HasAnnotationWithValue(annotations map[string]string, key, value string) bool {
```
preflightsiren, line 126: I renamed this method as this tests the `value` of the annotation, not just that it exists - happy for suggestions for method names :)

controllers/provisioners/eks/helpers.go
```diff
@@ -1173,3 +1173,40 @@ func (ctx *EksInstanceGroupContext) GetDesiredMixedInstancesPolicy(name string)
...
+	var (
+		instanceGroup = ctx.GetInstanceGroup()
+		state         = ctx.GetDiscoveredState()
+		configuration = instanceGroup.GetEKSConfiguration()
+	)
+	clusterVersion := state.GetClusterVersion()
+	annotations := instanceGroup.GetAnnotations()
+
+	var OSFamily string
+	if kubeprovider.HasAnnotation(annotations, OsFamilyAnnotation) {
```
preflightsiren, line 1199: previous version, now `HasAnnotationWithValue` caused some unexpected logic problems; I've added a new method to just detect the presence of a key in the annotation

preflightsiren:
@eytan-avisror @backjo this is ready for final review; I want to finish this off and get it ready to merge before improving/expanding it.

My thoughts for next steps are:
1. Support custom SSM Param path
2. Support custom Role assumption (primarily to support cross account ssm lookup)
3. `NodeUpgradeRequest` ala. #321

preflightsiren:
Do you prefer to have commits squashed before merge? so I can give it a nice commit message - or would you prefer to do it once merging the PR?

backjo:
Not sure if we have a policy for the project - I definitely prefer squashed commits before merge personally :)

docs/examples/EKS.md
```diff
@@ -177,3 +178,28 @@ spec:
...
+
+Amazon provides a EKS optimised AMIs for AmazonLinux2, Bottlerocket and Windows - https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html
+Instance-manager can periodically query the latest AMI ID from AWS SSM parameter store, and automatically update the image used in your instance groups, ensuring your nodes are always patched and kept up-to-date. This feature pairs well with customised upgrade strategies to ensure nodes are safely upgraded when and how you expect.
+
+note: The controller reconciles `InstanceGroups` every 10 hours by default, so changes may take up to 11 hours (sync period + SSM GetParameter cache TTL) to occur.
+Using the locking annotation `instancemgr.keikoproj.io/lock-upgrades` will allow more control of upgrades. Locking/Unlocking `InstanceGroups` will trigger a reconcile.
+
```
backjo, line 190: We should add the relevant SSM permissions required to the list of AWS permissions in INSTALL.md

controllers/provisioners/eks/helpers.go
```diff
@@ -1173,3 +1173,40 @@ func (ctx *EksInstanceGroupContext) GetDesiredMixedInstancesPolicy(name string)
 
 	return policy
 }
+
+func FilterSupportedArch(architectures []string) string {
```
eytan-avisror, line 1177: Is `FilterSupportedArch` only meant to check if a slice of architectures is in slice of `SupportedArchitectures` ?
Could we just use `common.StringSliceContains`, or a different generic function?
In the below code it seems we return the first architecture that we find a match for, is there a condition where there are multiple supported architectures?

preflightsiren: `FilterSupportedArch` returns the first match between the `SupportedArchitectures`
from my investigation of the aws api most instance types either return 
```
"arch": [
    "i386",
    "x86_64"
  ]
```
or
```
"arch": [
    "arm64"
  ]
```

since i386 isn't a supported arch, `common.StringSliceContains` would fail. Happy to use a different generic if one exists. If ever amazon provides an instance with multiple supported architectures, we would either need to explain the preference order, or provide some way of preferencing one.

eytan-avisror:
This is looking pretty good to me, @backjo do you have any additional comments?
Great work on this @preflightsiren

PR #332, opened 2021-09-15
fix: check architecture when considering similar instance types in the same family

### LINKED ISSUE #331: SubFamilyFlexible calculations should ensure that instance types have the same supported architectures.
BUG REPORT
**What happened**:
I configured an InstanceGroup to use spot instances with the SubFamilyFlexible option, with the main instance type being set to m6i.2xlarge. Instance Manager also included m6g.2xlarges in the ASG, which have a different supported architecture (arm64) then the main instance type (amd64)
Only instances with the same architecture should be used. 

**Environment**:

- Kubernetes version:

```bash
kubectl version -o yaml
```

- InstanceGroup status:

```bash
```

- controller logs:

```bash
```

backjo:
Closes #331 

### DIFF
--- a/controllers/providers/aws/aws.go
+++ b/controllers/providers/aws/aws.go
@@ -227,7 +227,7 @@ func GetInstanceArchitectures(typeInfo []*ec2.InstanceTypeInfo, instanceType str
 	}
 	var archs = []string{}
 	for _, s := range instanceTypeInfo.ProcessorInfo.SupportedArchitectures {
-		archs = append(archs, *s)
+		archs = append(archs, aws.StringValue(s))
 	}
 	sort.Strings(archs)
 	return archs
--- a/controllers/providers/aws/aws.go
+++ b/controllers/providers/aws/aws.go
@@ -18,6 +18,7 @@ package aws
 import (
 	"os"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


controllers/provisioners/eks/cloud.go
```diff
@@ -436,8 +438,19 @@ func subFamilyFlexiblePool(offerings []*ec2.InstanceTypeOffering, typeInfo []*ec
...
 					Weight: DefaultOfferingWeight,
 				}
+				supportedArchs = awsprovider.GetInstanceArchitectures(typeInfo, instanceType)
 			)
 
+			if len(desiredArchs) != len(supportedArchs) {
+				continue
+			}
+
+			for i, _ := range desiredArchs {
```
eytan-avisror, line 448: can we instead use `common.StringSliceContains`?
e.g. 
```go
if !common.StringSliceContains(supportedArchs, desiredArchs) {
    continue
}
```

controllers/providers/aws/aws.go
```diff
@@ -218,6 +219,20 @@ func GetInstanceFamily(instanceType string) string {
...
 
+func GetInstanceArchitectures(typeInfo []*ec2.InstanceTypeInfo, instanceType string) []string {
+
+	instanceTypeInfo := GetInstanceTypeInfo(typeInfo, instanceType)
+	if instanceTypeInfo == nil || instanceTypeInfo.ProcessorInfo == nil {
+		return []string{}
+	}
+	var archs = []string{}
+	for _, s := range instanceTypeInfo.ProcessorInfo.SupportedArchitectures {
+		archs = append(archs, *s)
```
eytan-avisror, line 230: Don't we need to dedup this list? if we go over multiple instances we might get same archs

eytan-avisror: Also, are there any arcithectures besides `i386 | x86_64 | arm64` ?
We can hardcode those and avoid using this.
Not sure if we expect this list to change

eytan-avisror: Oh nvm, it's for a single instance
Would be good if we can just use `aws.StringValue` instead of *s for nil safety (probably not a likely one though)

eytan-avisror (approved):
👍

PR #337, opened 2021-11-12
Stop passing in ClusterCA and Endpoint information for Windows nodes

backjo:
The Windows EKS bootstrap script currently contains a bug related to ServiceCidr calculations when passing in ClusterCA and APIEndpoint information - see https://github.com/awslabs/amazon-eks-ami/issues/805. Until the bootstrap script allows us to provide the ServiceCIDR, we should stop passing in the ClusterCA and Endpoint information to avoid incorrect CIDRs from being calculated.

### DIFF
--- a/controllers/instancegroup_controller.go
+++ b/controllers/instancegroup_controller.go
@@ -43,17 +43,18 @@ import (
 // InstanceGroupReconciler reconciles an InstanceGroup object
 type InstanceGroupReconciler struct {
 	client.Client
-	SpotRecommendationTime float64
-	ConfigNamespace        string
-	NodeRelabel            bool
-	Log                    logr.Logger
-	MaxParallel            int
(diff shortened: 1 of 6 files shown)


eytan-avisror:
Since it's a bug workaround, should we have a flag to control this behavior? e.g. `--disable-windows-cluster-ca-detection` or similar? it might be too specific, but that way we won't need another release when it's fixed and we end up giving more flexibility
WDYT

backjo:
@eytan-avisror it depends on how they fix it - if the fix is to require passing in a `ServiceCidr` parameter, we'd need another update to propagate that information. I'm fine flagging it for now.

backjo:
@eytan-avisror - added a flag for the controller to allow toggling injection.

eytan-avisror (approved):

PR #342, opened 2022-02-04
fix: don't set max pods above kubernetes recommended limit

backjo:

### DIFF
--- a/controllers/provisioners/eks/helpers.go
+++ b/controllers/provisioners/eks/helpers.go
@@ -489,6 +489,13 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
 	return labelList
 }
 
+func min(a, b int64) int64 {
+	if a < b {
+		return a
+	}
+	return b
+}
+
 func (ctx *EksInstanceGroupContext) GetComputedBootstrapOptions() *v1alpha1.BootstrapOptions {
 	var (
 		instanceGroup = ctx.GetInstanceGroup()
@@ -511,7 +518,8 @@ func (ctx *EksInstanceGroupContext) GetComputedBootstrapOptions() *v1alpha1.Boot
 			ipsPerInterface = 16 //Number of ips in a /28 block
 		}
 
-		maxPods = enis*((aws.Int64Value(instanceTypeNetworkInfo.Ipv4AddressesPerInterface)-1)*ipsPerInterface) + hostNetworkPods
+		// Don't set maxPods above Kubernetes-recommended 110 per node for large clusters.
+		maxPods = min(enis*((aws.Int64Value(instanceTypeNetworkInfo.Ipv4AddressesPerInterface)-1)*ipsPerInterface) + hostNetworkPods, 110)
 
 		if configuration.BootstrapOptions == nil {
 			return &v1alpha1.BootstrapOptions{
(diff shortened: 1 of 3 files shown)


controllers/provisioners/eks/helpers.go
```diff
@@ -511,7 +518,8 @@ func (ctx *EksInstanceGroupContext) GetComputedBootstrapOptions() *v1alpha1.Boot
 			ipsPerInterface = 16 //Number of ips in a /28 block
 		}
 
-		maxPods = enis*((aws.Int64Value(instanceTypeNetworkInfo.Ipv4AddressesPerInterface)-1)*ipsPerInterface) + hostNetworkPods
+		// Don't set maxPods above Kubernetes-recommended 110 per node for large clusters.
+		maxPods = min(enis*((aws.Int64Value(instanceTypeNetworkInfo.Ipv4AddressesPerInterface)-1)*ipsPerInterface) + hostNetworkPods, 110)
```
eytan-avisror, line 522: Should we make this configurable via `--max-pods-hard-limit`? if it's not set leave as is, so we can keep backwards comp. - not sure if anyone would be affected by the change. WDYT?

controllers/provisioners/eks/helpers.go
```diff
@@ -489,6 +489,13 @@ func (ctx *EksInstanceGroupContext) GetLabelList() []string {
 	return labelList
 }
 
+func min(a, b int64) int64 {
```
eytan-avisror, line 492: Probably move to common/utils.go - or you can also use math.Min() but it takes float. Whatever makes more sense 👍

controllers/provisioners/eks/helpers.go
backjo: I imagine usage here is probably pretty limited still, but open to changing it to the hard limit. I think making this new behavior default probably benefits everyone - right now, on the linux EKS-optimized AMIs, half of the memory on m6g.xlarge nodes with custom networking enabled is being allocated in KubeReserved.

eytan-avisror: I think if you define your own Kubelet limits it overrides that, but I see your point, I guess if it's unset on Kubelet config then it auto calculates it from max pods?

eytan-avisror: According to docs:
```


--kube-reserved mapStringString     Default: <None>
--
  | A set of <resource name>=<resource quantity> (e.g. cpu=200m,memory=500Mi,ephemeral-
storage=1Gi,pid='100') pairs that describe resources reserved for kubernetes system components. 
Currently cpu, memory and local ephemeral-storage  for root file system are supported.
 See  http://kubernetes.io/docs/user-guide/compute-resources for more detail.  
(DEPRECATED: This parameter should be set via the config file specified  by the Kubelet's 
--config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/
 for more information.)

```

I don't see any default, or explanation on impact of max-pods on kube reserved

backjo: See https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L436

backjo: That's the frustrating part - it's not _really_ documented.

eytan-avisror: I guess this will only affect people who use custom networking annotation, we can probably put a release not regarding this.
But this is a reasonable change.
Let's just move min to common and we should be good to merge 👍

eytan-avisror (approved):
🚀

PR #344, opened 2022-02-04
fix: Deduplicate Subnets

### LINKED ISSUE #339: Subnets field should deduplicate if duplicate values are provided
If someone mistakenly provides dupe values, e.g.:
```yaml
  subnets:
  - subnet-0310e3af3100225b9
  - subnet-0310e3af3100225b9
  - subnet-05c09777fce590171
```

The call will fail with:
```
"error": "provisioner eks reconcile failed: failed to update scaling group: ValidationError: Duplicate subnets specified: [subnet-0310e3af3100225b9]
```

We should consider having the controller dedupe values to avoid the error in this case and only use the two subnets.

Downside is that maybe we would like to explicitly provide this error if the user made a mistake and provided wrong subnets.

eytan-avisror:

Fixes #339 

This ignores duplicate values if provided in subnet field to avoid blocking reconcile.


backjo (approved):
LGTM. I wonder if it's worth emitting a warning-type event on the IG to give folks a heads up in case they did not intend to supply duplicate values.

eytan-avisror:
> LGTM. I wonder if it's worth emitting a warning-type event on the IG to give folks a heads up in case they did not intend to supply duplicate values.

Yep, our logger right now only has Info() and Error(), but allows for logging levels, we should do some work on logging.
For now emitting an info message.

PR #345, opened 2022-02-04
Enforce Restricted Fields usage when no default value

### LINKED ISSUE #341: Changing a boundary does not roll back changes on CRs
When using the admin config map, and changing a specific boundary, the changes does not seem to take effect.

e.g.
If `spec.eks.configuration.suspendProcesses` is a merge field, and a CR is using AZRebalance as suspendProcesses, once the boundary is changed to restricted (and there is no default value for suspendProcesses), the ASG remains with AZRebalance enabled.

Ideally, once the boundary is changed, the change should be reconciled across all Custom Resources.

eytan-avisror:

Fixes #341 

This fixes a gap in our admin configuration where we allow CRs to use restricted fields if there are no defaults.
This change sets field to Nil if there is no default value provided and field is a restricted field.


controllers/provisioners/config.go
```diff
@@ -189,6 +190,12 @@ func (c *ProvisionerConfiguration) setRestrictedFields(unstructuredInstanceGroup
 			if err != nil {
 				errors.Wrap(err, "failed to set nested field")
 			}
+		} else {
+			// if no default value exist, make sure to ignore CR value for restricted fields
+			err := unstructured.SetNestedField(unstructuredInstanceGroup, nil, path...)
+			if err != nil {
+				errors.Wrap(err, "failed to set nested field to nil")
```
eytan-avisror, line 197: @backjo I just noticed we are not returning any of these errors here.. do you remember if we did this intentionally? Thinking we should return these errors to caller WDYT

backjo: :+1:  I don't think it was intentional

backjo (approved):

PR #346, opened 2022-02-09
bug fix for launchTemplate loop reconcile

### LINKED ISSUE #340: Endless LaunchTemplate creation loop when reconcile errors occur
When reconcile errors occur with LaunchTemplate, it seems we end up in an endless loop of creating launch templates.
Eventually the limit is reached and manual intervention is required to delete.

We should:
- Make sure the controller never leaks launch templates (i.e. delete older launch templates before creating new ones)
- Make sure when reconcile errors occur we do not keep creating new launch templates.

Logs:
```
2022-01-28T18:09:55.836Z	INFO	v1alpha1	state transition occured	{"instancegroup": "instance-manager/nodes", "state": "InitUpdate", "previousState": "Init"}
2022-01-28T18:09:55.836Z	INFO	v1alpha1	state transition occured	{"instancegroup": "instance-manager/nodes", "state": "ReconcileModifying", "previousState": "InitUpdate"}
2022-01-28T18:09:55.836Z	INFO	scaling	detected drift	{"reason": "launchtemplate does not exist", "instancegroup": "instance-manager/nodes"}
2022-01-28T18:09:55.944Z	DEBUG	aws-provider	AWS API call	{"cacheHit": false, "service": "ec2", "operation": "CreateLaunchTemplate"}
2022-01-28T18:09:55.945Z	INFO	controllers.instancegroup.eks	node rotation required	{"instancegroup": "instance-manager/nodes", "scalingconfig": "instance-manager-nodes-20220128180955"}
2022-01-28T18:09:56.164Z	DEBUG	aws-provider	AWS API call	{"cacheHit": false, "service": "autoscaling", "operation": "UpdateAutoScalingGroup"}
2022-01-28T18:09:56.164Z	INFO	v1alpha1	state transition occured	{"instancegroup": "instance-manager/nodes", "state": "Error", "previousState": "ReconcileModifying"}
2022-01-28T18:09:56.165Z	INFO	controllers.instancegroup	patching resource status	{"instancegroup": "instance-manager/nodes", "patch": "{\"status\":{\"activeLaunchTemplateName\":\"instance-manager-nodes-20220128180955\"}}", "resourceVersion": "7699932"}
2022-01-28T18:09:56.177Z	ERROR	controller-runtime.manager.controller.instancegroup	Reconciler error	{"reconciler group": "instancemgr.keikoproj.io", "reconciler kind": "InstanceGroup", "name": "nodes", "namespace": "instance-manager", "error": "provisioner eks reconcile failed: failed to update scaling group: ValidationError: Cannot configure weights because instance type [t2.large] is not in the overrides but is currently running in the group. Add [t2.large] to the overrides, specify a weight, and try again.\n\tstatus code: 400, request id: 947c73a5-d361-423c-ba0c-6f0c06410bd4", "errorVerbose": "ValidationError: Cannot configure weights because instance type [t2.large] is not in the overrides but is currently running in the group. Add [t2.large] to the overrides, specify a weight, and try again.\n\tstatus code: 400, request id: 947c73a5-d361-423c-ba0c-6f0c06410bd4\nfailed to update scaling group\ngithub.com/keikoproj/instance-manager/controllers/provisioners/eks.(*EksInstanceGroupContext).Update
```

In this case we tried switching from t2 to m5 with mixed instances - this is another possible bug we should address

sahilbadla:
Signed-off-by: sbadla1 <sahil_badla@intuit.com>

closes #340 

Endless LaunchTemplate creation loop happens when swtiching from launchConfig to LaunchTemplate.
Fixed the launchTemplate discovery logic to pass latest template config instead of ASG config(that still has launchConfig)

### DIFF
--- a/controllers/provisioners/eks/cloud.go
+++ b/controllers/provisioners/eks/cloud.go
@@ -238,7 +238,7 @@ func (ctx *EksInstanceGroupContext) CloudDiscovery() error {
 
 	if spec.IsLaunchTemplate() {
 		state.ScalingConfiguration, err = scaling.NewLaunchTemplate(instanceGroup.NamespacedName(), ctx.AwsWorker, &scaling.DiscoverConfigurationInput{
-			ScalingGroup: targetScalingGroup,
+			TargetConfigName: status.GetActiveLaunchTemplateName(),
 		})
 		if err != nil {
 			return errors.Wrap(err, "failed to discover launch templates")


controllers/provisioners/eks/cloud.go
```diff
@@ -238,7 +238,7 @@ func (ctx *EksInstanceGroupContext) CloudDiscovery() error {
 
 	if spec.IsLaunchTemplate() {
 		state.ScalingConfiguration, err = scaling.NewLaunchTemplate(instanceGroup.NamespacedName(), ctx.AwsWorker, &scaling.DiscoverConfigurationInput{
-			ScalingGroup: targetScalingGroup,
```
eytan-avisror, line 241: I think the current implementation discovers the active launch template by the scaling group name, can you look inside NewLaunchTemplate and see if there is a reason this is not happening?

eytan-avisror: Looks like we are not setting TargetConfigName inside Discover:
https://github.com/sahilbadla/instance-manager/blob/58ad1a253a80900c2a1047c5809ca1d94e300d00/controllers/provisioners/eks/scaling/launchtemplate.go#L58

Can you try that instead? 
That way we don't rely on status field, but rather discover the active Launch Template name each time from the scaling group name.

eytan-avisror (approved):

PR #351, opened 2022-03-03
feat: allow resolving EKS AMIs by version/slug for better readability…

backjo:
… and portability


eytan-avisror:
@backjo this is awesome, love this abstraction, will review shortly

controllers/providers/aws/ssm.go
```diff
@@ -60,8 +65,16 @@ func GetAwsSsmClient(region string, cacheCfg *cache.Config, maxRetries int, coll
 }
 
 func (w *AwsWorker) GetEksLatestAmi(OSFamily string, arch string, kubernetesVersion string) (string, error) {
+	return w.GetEksSSMAmi(OSFamily, arch, kubernetesVersion, LatestIdentifiers[OSFamily])
+}
+
+func (w *AwsWorker) GetEksSSMAmi(OSFamily string, arch string, kubernetesVersion string, ssmId string) (string, error) {
```
eytan-avisror, line 71: nit: Name is a bit weird, GetEksSsmAmi vs GetEKSSSMAMI vs GetEksSSMAmi - not sure what is the right golang convention but looks like in other methods we dont upcase the abbreviations.
Not really important 🤣

controllers/providers/aws/ssm.go
```diff
@@ -60,8 +65,16 @@ func GetAwsSsmClient(region string, cacheCfg *cache.Config, maxRetries int, coll
...
+	var inputString = aws.String(fmt.Sprintf(EksAmis[OSFamily][arch], kubernetesVersion, ssmId))
+	if OSFamily == "windows" {
+		inputString = aws.String(fmt.Sprintf(EksAmis[OSFamily][arch], ssmId, kubernetesVersion))
```
eytan-avisror, line 74: Do we validate the annotation value OSFamily / Arch upstream prior to this? Can you make sure that invalid annotation value to dont result in exception due to map lookup?

controllers/provisioners/eks/cloud.go
```diff
@@ -186,6 +186,16 @@ func (ctx *EksInstanceGroupContext) CloudDiscovery() error {
 		ctx.Log.V(4).Info("Updating Image ID with latest", "ami_id", latestAmiId)
 	}
 
+	if strings.HasPrefix(configuration.Image, v1alpha1.ImageSSMPrefix) {
+		ssmKey := strings.TrimPrefix(configuration.Image, v1alpha1.ImageSSMPrefix)
+		amiId, err := ctx.GetEksSSMAmi(&ssmKey)
```
eytan-avisror, line 191: Does ssmKey need to be a ptr here?

backjo: Nah, we can change it

docs/examples/EKS.md
```diff
@@ -202,4 +202,30 @@ spec:
...
+  name: hello-world
+  namespace: instance-manager
+  annotations:
+    instancemgr.keikoproj.io/os-family: "amazonlinux2" #Default if not provided
+spec:
+  strategy: <...>
+  provisioner: eks
+  eks:
+    configuration:
+      image: ssm://amazon-eks-node-1.18-v20220226 # resolves to value in /aws/service/eks/optimized-ami/1.18/amazon-linux-2/amazon-eks-node-1.18-v20220226/image_id
```
eytan-avisror, line 226: So the difference essentially between `latest` and `ssm://amazon-eks-node-1.18-v20220226` is that you get control over the `/amazon-eks-node-1.18-v20220226/` section of the URI, where `latest` would probably get a 1.22 or newer version in that case?
Should we also create a bug for the `latest` case? it seems latest should also pin to the cluster version

eytan-avisror: Also what we'll get by fixing the `latest` pinning case is once the cluster is upgraded to a newer version, it will immediately trigger a node rotation

backjo: Latest is already looking up the cluster k8s version, and selecting the appropriate AMI.

controllers/provisioners/eks/update_test.go
```diff
@@ -738,9 +745,10 @@ func TestUpdateWithLatestAmiID(t *testing.T) {
 	}
 
 	// Setup Latest AMI
-	ig.GetEKSConfiguration().Image = "latest"
-	ssmMock.latestAMI = testLatestAmiID
-
+	ssmMock.parameterMap = map[string]string{
+		"/aws/service/eks/optimized-ami/1.18/amazon-linux-2/recommended/image_id":                    testLatestAmiID,
+		"/aws/service/eks/optimized-ami/1.18/amazon-linux-2/amazon-eks-node-1.18-v20220226/image_id": testCustomAmi,
+	}
```
eytan-avisror, line 751: Do we want to add cases for Windows/Bottlerocket, or do you feel it would be redundant?
Might be nice since it if this breaks somehow we would be in trouble.

Same for the other tests

docs/examples/EKS.md
eytan-avisror: Oh OK, so essentially why we want this feature is to be able to control `-v20220226` ?

eytan-avisror: Or if you don't want your nodes to move with cluster

backjo: Yes, the former.

eytan-avisror (approved):
LG!

PR #355, opened 2022-04-05
fix: use non-deprecated CRD api

backjo:

### DIFF
--- a/controllers/providers/kubernetes/utils.go
+++ b/controllers/providers/kubernetes/utils.go
@@ -218,7 +218,7 @@ func GetKubernetesLocalConfig() (*rest.Config, error) {
 }
 
 func CRDExists(kubeClient dynamic.Interface, name string) bool {
-	CRDSchema := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1beta1", Resource: "customresourcedefinitions"}
+	CRDSchema := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1", Resource: "customresourcedefinitions"}
 	_, err := kubeClient.Resource(CRDSchema).Get(context.Background(), name, metav1.GetOptions{})
 	if err != nil {
 		return false
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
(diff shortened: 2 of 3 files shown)


controllers/providers/kubernetes/utils.go
```diff
@@ -218,7 +218,7 @@ func GetKubernetesLocalConfig() (*rest.Config, error) {
 }
 
 func CRDExists(kubeClient dynamic.Interface, name string) bool {
-	CRDSchema := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1beta1", Resource: "customresourcedefinitions"}
+	CRDSchema := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1", Resource: "customresourcedefinitions"}
```
eytan-avisror, line 221: oh wow, nice catch, didn't see this one when evaluating deprecated APIs
This would probably not be backwards compatible though, would it? i.e. what happens to people who are still running 1.19 once we change this.

eytan-avisror: (I believe this deprecation happens in 1.22 which is pretty new)

backjo: V1 went GA in 1.16 IIRC

backjo: Yep, per https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/ - V1 has been available since 1.16

eytan-avisror: CRDExists is used by upgrade stratgegy for crd strategy if you want to use some CRD to update.
In this case, if you were using upgrade-manager's RollingUpgrade CRD, but you are still using v1beta1 version of it, you would have a problem right?

eytan-avisror: Thats fine then, we should just probably add a release note on this being a potentially breaking change

backjo: You shouldn't - per Kubernetes docs 

```
For example, suppose there are two API versions, v1 and v1beta1, for the same resource. If you originally created an object using the v1beta1 version of its API, you can later read, update, or delete that object using either the v1beta1 or the v1 API version.
```

eytan-avisror: Right, so I guess it's a storedVersion for CRD type

eytan-avisror (approved):

PR #357, opened 2022-04-12
feat: use IMDS for autoscaling information

backjo:


eytan-avisror:
Awesome, this makes warm pool support much easier..
Did you get a chance to run this and make sure it works? Not sure if our functional testing covers warm pool or not

backjo:
> Awesome, this makes warm pool support much easier.. Did you get a chance to run this and make sure it works? Not sure if our functional testing covers warm pool or not

I haven't - I believe we have a warm pool TC in the functional tests - was planning to use that to verify. I validated that the IMDS endpoint was returning the expected values on a  test instance

eytan-avisror:
> > Awesome, this makes warm pool support much easier.. Did you get a chance to run this and make sure it works? Not sure if our functional testing covers warm pool or not
> 
> I haven't - I believe we have a warm pool TC in the functional tests - was planning to use that to verify. I validated that the IMDS endpoint was returning the expected values on a test instance

Yes, looks like we do.. would probably good to do a quick manual test since bdd tests basic CRUD, but I believe if user data fails for any reason node will not join and test will fail, so it should be sufficient

eytan-avisror (approved):
LGTM

PR #366, opened 2022-04-28
added drift detection logic to notify drifted IGs

nrosenberg1:


controllers/verticalscalingpolicy_controller.go
```diff
@@ -54,6 +53,7 @@ type VerticalScalingPolicyReconciler struct {
 func (r *VerticalScalingPolicyReconciler) Reconcile(ctxt context.Context, req ctrl.Request) (ctrl.Result, error) {
 	_ = context.Background()
 	_ = r.Log.WithValues("verticalscalingpolicy", req.NamespacedName)
+	var driftedTargets []string
```
nvandanapu, line 56: Could you make this a map to avoid duplicates? Although it's not possible to have duplicates now, we can make this future proofed

controllers/verticalscalingpolicy_controller.go
```diff
@@ -351,14 +352,14 @@ func instanceFamilyExists(family string, instanceTypesInfo []*ec2.InstanceTypeIn
...
+func (r *VerticalScalingPolicyReconciler) NotifyTargets(targets []string) {
 	for _, target := range targets {
 		r.ManagerContext.InstanceGroupEvents <- event.GenericEvent{
 			Object: &metav1.PartialObjectMetadata{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      target.Name,
-					Namespace: target.Namespace,
+					// Target name is in format my-namespace/my-ig
+					Namespace: strings.Split(target, "/")[0],
+					Name:      strings.Split(target, "/")[1],
```
nvandanapu, line 362: In future, we can use regex match to get namespace and ig names from target string to avoid array out of bounds exception

controllers/verticalscalingpolicy_controller.go
nrosenberg1: Sure I will make that change, good call

controllers/verticalscalingpolicy_controller.go
nrosenberg1: Yeah, I will add a TODO note here

nvandanapu (approved):

PR #372, opened 2022-05-23
fix: check for role attached to instance profile and create if it doe…

### LINKED ISSUE #371: Instance profile is stuck in bad state if error occurs when initial role is attached
BUG
**What happened**:
An Instance profile was created by IM, but it had no role attached. It is stuck in this state. 
IM should reconcile this error correctly, and attach a role to the profile if one does not exist on subsequent checks.

**Environment**:

- Kubernetes version:

```bash
kubectl version -o yaml
```

- InstanceGroup status:

```bash
```

- controller logs:

```bash
```

backjo:
Fixes #371


controllers/providers/aws/iam.go
```diff
@@ -224,10 +229,9 @@ func (w *AwsWorker) CreateScalingGroupRole(name string) (*iam.Role, *iam.Instanc
 					return createdRole, createdProfile, errors.Wrap(err, "failed to attach instance-profile")
 				}
 			}
+		} else {
```
eytan-avisror, line 232: Is there an indentation problem from line 226?

eytan-avisror (approved):

Issue #379, opened 2023-04-01
not honoring gp3 volume iops and throughput values


vgunapati:
**Is this a BUG REPORT or FEATURE REQUEST?**:

**What happened**:

### Describe the bug

As part of gp3 testing, added rootVolume fields for gp3 volumeType and configuration. https://github.intuit.com/gitops/iksm-cluster-gitops/pull/4314/files

Instance manager created volume with `Type:gp3` but not  honoring `iops` and `throughput` values.

![image](https://user-images.githubusercontent.com/1102401/229255778-952dadfc-9f69-4db9-8cc3-9093525fab71.png)

IntanceGroup Spec:
```
apiVersion: instancemgr.keikoproj.io/v1alpha1
kind: InstanceGroup
metadata:
  annotations:
    instancemgr.keikoproj.io/irsa-enabled: "true"
  creationTimestamp: "2020-12-01T12:46:05Z"
  finalizers:
  - finalizer.instancegroups.keikoproj.io
  generation: 38
  name: nodes
  namespace: instance-manager
spec:
  eks:
    configuration:
      instanceType: m5.xlarge
      tags:
      - key: k8s.io/cluster-autoscaler/node-template/resources/ephemeral-storage
        value: 435Gi
      volumes:
      - iops: 3500
        name: /dev/xvda
        size: 512
        throughput: 250
        type: gp3
    maxSize: 50
    minSize: 3
  provisioner: eks
```

### To Reproduce

1. TBD

### Root Cause

Describe how this bug was introduced and how similar bugs could be prevented in the future.

### Additional context

> **NOTE:** Please be sure to add the appropriate `area/*` and priority labels. Please also set the Projects to include `IKS Team FYxx-Qx` to indicate when this work should be done.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Kubernetes version:

```bash
kubectl version -o yaml
```

**Other debugging information (if applicable)**:

- InstanceGroup status:

```bash
kubectl describe instancegroup <ig-name>
```

- controller logs:

```bash
kubectl logs <instance-manager pod>
```

PR #380, opened 2023-04-01
Fix: Not honoring gp3 volume iops and throughput values

vgunapati:
Fixes https://github.com/keikoproj/instance-manager/issues/379


controllers/providers/aws/constructors.go
```diff
@@ -40,20 +38,22 @@ func (w *AwsWorker) GetAutoScalingBasicBlockDevice(name, volType, snapshot strin
 	if encrypt != nil {
 		device.Ebs.Encrypted = encrypt
 	}
-	if iops != 0 && strings.EqualFold(volType, "io1") {
+	if iops != 0 && common.ContainsEqualFold(AllowedVolumeTypesWithProvisionedIOPS, volType) {
 		device.Ebs.Iops = aws.Int64(iops)
 	}
+	if throughput != 0 && common.ContainsEqualFold(AllowedVolumeTypesWithProvisionedThroughput, volType) {
```
eytan-avisror, line 44: Is there a default on the SDK side if it's set to 0?

eytan-avisror (approved):
LGTM 
We can do a release if you need it and the functional-test is passing

controllers/providers/aws/constructors.go
vgunapati: old SDK did not even have the option to configure Throughput, So I have to bump up to the new version, Tested `0` on gp3 for ProvisionedThroughput it was treated as `nil`, and the default value is set in both the launch template and launch configuration

eytan-avisror: 👌🏻

vgunapati:
Looks like [functional-test](https://github.com/keikoproj/instance-manager/actions/runs/2895189866) is broken for a while, Do we have a doc on how to set this up if the cluster is deleted

eytan-avisror:
No doc, but it pretty much needs a vanilla EKS cluster + upgrade manager

vgunapati:
I don't have access to the settings page to update Access keys, I will follow up with the team

PR #385, opened 2023-08-22
Make launchtemplates as a default scaling group configuration, instead of launchconfigurations.

shreyas-badiger:
The configurations for the AutoscalingGroups could be specified in either a launchConfiguration or a launchTemplate. launchtemplates has several benefits over launchconfigurations:
- ability to have increased number of scaling groups in a cluster. 
- support for mixed instance type in a ASG.
- support for newer instance type in ASG.

[aws blog-post](https://aws.amazon.com/blogs/compute/amazon-ec2-auto-scaling-will-no-longer-add-support-for-new-ec2-features-to-launch-configurations/)

>  AWS released [launch templates](https://aws.amazon.com/about-aws/whats-new/2017/11/introducing-launch-templates-for-amazon-ec2-instances/), the successor of launch configurations, as a way to streamline and simplify the launch process for [Auto Scaling](https://aws.amazon.com/ec2/autoscaling/), [Spot Fleet](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-fleet.html), [Amazon EC2 Spot Instances](https://aws.amazon.com/ec2/spot/?cards.sort-by=item.additionalFields.startDateTime&cards.sort-order=asc&trk=ps_a134p000006vwG3AAI&trkCampaign=acq_paid_search_brand&sc_channel=PS&sc_campaign=acquisition_US&sc_publisher=Google&sc_category=Cloud%20Computing&sc_country=US&sc_geo=NAMER&sc_outcome=acq&sc_detail=ec2%20spot&sc_content=EC2%20Spot_e&sc_matchtype=e&sc_segment=517520538473&sc_medium=ACQ-P%7CPS-GO%7CBrand%7CDesktop%7CSU%7CCloud%20Computing%7CEC2%20Spot%7CUS%7CEN%7CText&s_kwcid=AL!4422!3!517520538473!e!!g!!ec2%20spot&ef_id=CjwKCAjwmK6IBhBqEiwAocMc8nuPFmpaTxCYXHIx2LzTmn2IjmsT8GUoWO53MOcvmJVS6wgVOUqxExoCL00QAvD_BwE:G:s&s_kwcid=AL!4422!3!517520538473!e!!g!!ec2%20spot), and [On-Demand Instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-on-demand-instances.html). Launch templates define the steps required to create an instance, by capturing instance parameters in a resource that can be used across multiple services. [Launch configurations](https://docs.aws.amazon.com/autoscaling/ec2/userguide/LaunchConfiguration.html) have continued to live alongside launch templates but haven’t benefitted from all of the features we’ve added to launch templates.
        
> Today, AWS is recommending that customers using [launch configurations](https://docs.aws.amazon.com/autoscaling/ec2/userguide/LaunchConfiguration.html) migrate to [launch templates](https://docs.aws.amazon.com/autoscaling/ec2/userguide/LaunchTemplates.html). We will continue to support and maintain launch configurations, but we will not be adding any new features to them. We will focus on adding new EC2 features to launch templates only. You can continue using launch configurations, and AWS is committed to supporting applications you have already built using them, but in order for you to take advantage of our most recent and upcoming releases, a migration to launch templates is recommended. Additionally, we plan to no longer support new instance types with launch configurations by the end of 2022. Our goal is to have all customers moved over to launch templates by then.

Additionally, launchconfigurations is being slowly phased out.
Screenshot 2023-08-22 at 3 18 17 PM


ZihanJiang96 (approved):

controllers/instancegroup_controller.go
```diff
@@ -201,6 +202,10 @@ func (r *InstanceGroupReconciler) Reconcile(ctxt context.Context, req ctrl.Reque
 		return ctrl.Result{}, errors.Wrapf(err, "provisioner %v reconcile failed", provisionerKind)
 	}
 
+	if r.SetScalingConfigurationToLaunchConfig {
+		input.InstanceGroup.Spec.EKSSpec.Type = v1alpha1.LaunchConfiguration
```
eytan-avisror, line 206: This flag is a bit strange, when set it forces specs to use launch template instead of launch configuration even though the spec requests launch config explicitly.
I think it makes more sense to introduce a flag that modifies the unspecified default.
That way resources can either omit the field and get the flag-controlled default, or explicitly declare a configuration which should be honored.

With the flag as is, you force all resources to switch and not sure it makes sense

eytan-avisror (changes requested):

kevdowney (approved):
LGTM

eytan-avisror (approved):

vgunapati:
Update the release notes with this change and clearly point out that we need to add additional permissions for Launch templates if missing, so that the controller can make seamless transition from launch configuration  to Launch templates

tekenstam (commented):
LGTM

PR #409, opened 2023-10-13
fix incorrect counter name

ZihanJiang96:


shreyas-badiger (approved):

yuxuanzhu0214 (commented):
Might also need to pass in the `reason` label since `failureCounter` has that when initialized.

Issue #445, opened 2024-08-29
Create an informer in instance manager to cache custom resources changes


yuxuanzhu0214:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

**What happened**:
We want to have a controller that watches for changes in instance manager CR. One way is to rely on API server and poll periodically for all instance manager CR, but this increases the load on API server. The other way is a informer that caches the changes of instance manager CR and this event driven method reduces load on API server.

**What you expected to happen**:
The informer can be generated using the sample script `update-codegen.sh` in this repo: https://github.com/kubernetes/sample-controller/blob/master/hack/update-codegen.sh

PR #448, opened 2024-09-09
add informer client code to instance manager generated by codegen

yuxuanzhu0214:
Added informer client code using codegen so component can import client informer code from instance manager.

fixes: https://github.com/keikoproj/instance-manager/issues/445


controllers/interface.go
```diff
@@ -1,7 +1,7 @@
 package controllers
 
 import (
-	v1alpha "github.com/keikoproj/instance-manager/api/v1alpha1"
+	v1alpha "github.com/keikoproj/instance-manager/api/instancemgr/v1alpha1"
```
shreyas-badiger, line 4: why change the directory structure?

ZihanJiang96 (approved):
lgtm

controllers/interface.go
ZihanJiang96: because we are using the open source code generator to generate the clientset/informer, and they are expecting the apis should have the following directory structure:
[api(s)/$group/$version](https://github.com/kubernetes/code-generator/blob/release-1.26/generate-groups.sh#L67-L69)

most of the open source k8s controllers are following this pattern

shreyas-badiger (approved):

